PR #118, opened 2018-09-27
(read (get-line!)) does not return the input line

luqui:
A quickcheck run failed, so I cast it into stone.  This is a bug, not a PR.  What's the process with this kinda thing?  "Esp since 'includes a failing test case' is such a common example for us..."


luqui:
Eclipsed by smarter fix of issue 16.

PR #144, opened 2018-10-12
Some changes to docs

jameshaydon:


docs/source/guide/Basics.lrad
```diff
@@ -1,9 +1,11 @@
...
 
-Though `radicle` is designed with a particular use-case in mind--namely, 
-defining and interacting with the Open Source Network chains--it is ultimately
-a general purpose language, quite similar to Scheme or Clojure. To effectivel
-use it, it is helpful to learn the basics of it independent of its application.
+Though `radicle` is designed with a particular use-case in mind--namely,
+defining and interacting with the Open Source Network chains--it is ultimately a
+general purpose language, quite similar to
+[Scheme](en.wikipedia.org/wiki/Scheme_%28programming_language%29) or
+[Clojure](clojure.org/). To effectivel use it, it is helpful to learn the basics
```
jkarni: Typo (`effectively`).

docs/source/guide/Basics.lrad
```diff
@@ -31,10 +32,55 @@ Strings are enclosed in double quotes:
 "this is a string"
 ```
 
+### Vectors
+
+For talking about sequences of values `radicle` has vectors. Literal notation
```
jkarni: Hm, the phrasing "For talking about sequences" is a little strange. How about something like "`radicle`'s *vectors* are an array-like container for values"?

docs/source/guide/Basics.lrad
```diff
@@ -31,10 +32,55 @@ Strings are enclosed in double quotes:
...
+["this" "is" "a" "vector"]
+```
+
+You can access any of the elements in the vector using `nth`:
+
+```radicle
+(nth 2 ["start at zero!" "one" "two" "three"]) ;; ==> "two"
+```
+
+But be careful: if the size of the argument isn't less than (<) the size of the
```
jkarni: (`<`)

docs/source/guide/Basics.lrad
```diff
@@ -31,10 +32,55 @@ Strings are enclosed in double quotes:
...
 ### Lists
 
-Lists are one of the fundamental data structures of functional
-programming. 
+Lists are one of the fundamental data structures of functional programming.
+
+The literal syntax for lists is just to write the elements down surrounded by
+parenthesis. In fact we've been using this notation up till now for all the rest
+of the code! So when you write down a list, the `radicle` interpreter thinks you
+want to evaluate something, so if the first element isn't a function (or some
```
jkarni: Hm, this is confusing. The `So` suggests some logical connection between us using it for all the rest of the code and the evaluation semantics of lists; there of course is such a connection, but from just this text it's not obvious why.

How about:

> Just like vectors are enclosed in square brackets, lists are enclosed in parentheses. 
> But as you may have already noticed, parentheses are also used for function application. Literal lists are *evaluated* as the application of their head to the rest of the list. (This connection between how code and data is represented forms the basis for *homoiconicity*, which we will talk more about later.)

PR #156, opened 2018-10-18
Added some From/ToRad instances and primfns.

jameshaydon:
Non hash/merkle/authenticated stuff of #141.

### DIFF
--- a/src/Radicle/Internal/Core.hs
+++ b/src/Radicle/Internal/Core.hs
@@ -19,7 +19,8 @@ import qualified Data.IntMap as IntMap
 import           Data.List.NonEmpty (NonEmpty)
 import qualified Data.List.NonEmpty as NonEmpty
 import qualified Data.Map as Map
-import           Data.Scientific (Scientific, floatingOrInteger)
+import           Data.Scientific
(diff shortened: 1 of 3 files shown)


src/Radicle/Internal/Core.hs
```diff
@@ -544,8 +555,13 @@ class ToRad t a where
 
 instance CPA t => ToRad t () where
     toRad _ = Vec Empty
+instance CPA t => ToRad t Bool where
+    toRad = Boolean
 instance (CPA t, ToRad t a, ToRad t b) => ToRad t (a,b) where
     toRad (x,y) = Vec $ toRad x :<| toRad y :<| Empty
+instance (CPA t, ToRad t a) => ToRad t (Maybe a) where
+    toRad Nothing  = Keyword (Ident "Nothing")
+    toRad (Just x) = Vec $ Keyword (Ident "Just") :<| toRad x :<| Empty
```
jkarni: Can you add `testForType` for a `Maybe` type [here](https://github.com/oscoin/radicle/blob/master/test/spec/Radicle/Tests.hs#L663)?

PR #158, opened 2018-10-18
String literals are parsed/printed with escape sequences

### LINKED ISSUE #126: Pretty . Parse not always reversible
I've run into some cases where having a newline `\n` inside a string causes the Pretty instance to yield code that modifies the whitespace inside of those strings. When parsing that back in, you get a different AST than the original.

jameshaydon:
And they work as defined in the Haskell report.

Fixes #126.

### DIFF
--- a/src/Radicle/Internal/Parse.hs
+++ b/src/Radicle/Internal/Parse.hs
@@ -23,7 +23,7 @@ import           Text.Megaparsec
                  , (<?>)
                  )
 import qualified Text.Megaparsec as M
-import           Text.Megaparsec.Char (char, satisfy, space1)
+import           Text.Megaparsec.Char (char, satisfy, space1, string)
 import qualified Text.Megaparsec.Char.Lexer as L
 import qualified Text.Megaparsec.Error as Par
 
@@ -47,7 +47,6 @@ tag v = do
     pos <- getPosition -- N.B. in megaparsec 7 this is called getSourcePos
     pure $ Ann.Annotated (Ann.WithPos (Ann.SrcPos pos) v)
 
-
 symbol :: Text -> Parser Text
 symbol = L.symbol spaceConsumer
 
@@ -66,9 +65,12 @@ bracesP = inside "{" "}"
 sqBracketsP :: Parser a -> Parser a
 sqBracketsP = inside "[" "]"
 
+-- Parsing of string literals handles escape sequences just like Haskell does.
 stringLiteralP :: VParser
-stringLiteralP = lexeme $
-    tag =<< StringF . toS <$> (char '"' >> manyTill L.charLiteral (char '"'))
+stringLiteralP = lexeme $ tag =<< StringF . toS <$> escaptedString
+  where
+    escaptedString = catMaybes <$> (char '"' >> manyTill ch (char '"'))
+    ch = (Just <$> L.charLiteral) <|> (Nothing <$ string "\\&")
 
 boolLiteralP :: VParser
 boolLiteralP = lexeme $ tag =<< BooleanF <$> (char '#' >>
--- a/src/Radicle/Internal/Parse.hs
+++ b/src/Radicle/Internal/Parse.hs
@@ -67,7 +67,7 @@ sqBracketsP = inside "[" "]"
 
 -- Parsing of string literals handles escape sequences just like Haskell does.
 stringLiteralP :: VParser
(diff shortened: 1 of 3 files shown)


src/Radicle/Internal/Parse.hs
```diff
@@ -66,9 +65,12 @@ bracesP = inside "{" "}"
 sqBracketsP :: Parser a -> Parser a
 sqBracketsP = inside "[" "]"
 
+-- Parsing of string literals handles escape sequences just like Haskell does.
 stringLiteralP :: VParser
-stringLiteralP = lexeme $
-    tag =<< StringF . toS <$> (char '"' >> manyTill L.charLiteral (char '"'))
+stringLiteralP = lexeme $ tag =<< StringF . toS <$> escaptedString
+  where
+    escaptedString = catMaybes <$> (char '"' >> manyTill ch (char '"'))
```
jkarni: escaptedT

jameshaydon: image

jkarni:
```suggestion
    escapedString = catMaybes <$> (char '"' >> manyTill ch (char '"'))
```

jameshaydon: From what I'm seeing in the UI you didn't change the call-site, just the define-site?

src/Radicle/Internal/Parse.hs
```diff
@@ -66,9 +65,12 @@ bracesP = inside "{" "}"
 sqBracketsP :: Parser a -> Parser a
 sqBracketsP = inside "[" "]"
 
+-- Parsing of string literals handles escape sequences just like Haskell does.
 stringLiteralP :: VParser
-stringLiteralP = lexeme $
-    tag =<< StringF . toS <$> (char '"' >> manyTill L.charLiteral (char '"'))
+stringLiteralP = lexeme $ tag =<< StringF . toS <$> escaptedString
```
jkarni:
```suggestion
stringLiteralP = lexeme $ tag =<< StringF . toS <$> escapedString
```

src/Radicle/Internal/Parse.hs
jkarni: Wanna make a suggestion on my suggestion?

jameshaydon: Lol I didn't try

src/Radicle/Internal/Pretty.hs
```diff
@@ -50,7 +50,8 @@ instance (Copointed t, Ann.Annotation t) => Pretty (Ann.Annotated t ValueF) wher
                         , sep $ pretty <$> toList vals
                         ])
       where
-        escapeStr = T.replace "\"" "\\\"" . T.replace "\\" "\\\\"
+        -- We print string literals escaped just like Haskell does.
```
jameshaydon:
```suggestion
        -- Radicle prints string literals escaped just like Haskell does.
```

jameshaydon: I suggested to myself to change this comment.

jameshaydon:
```suggestion
        -- We print string literals escaped just like Haskell does.
```

jameshaydon: No idea what that just did, seems you can just add another suggestion instead. But it can't be applied until the first one is?

jameshaydon: Ah it was invalid because no changes.
```suggestion
        -- Everyone prints string literals escaped just like Haskell does.
```

src/Radicle/Internal/Parse.hs
```diff
@@ -66,9 +65,12 @@ bracesP = inside "{" "}"
 sqBracketsP :: Parser a -> Parser a
 sqBracketsP = inside "[" "]"
 
+-- Parsing of string literals handles escape sequences just like Haskell does.
 stringLiteralP :: VParser
-stringLiteralP = lexeme $
-    tag =<< StringF . toS <$> (char '"' >> manyTill L.charLiteral (char '"'))
+stringLiteralP = lexeme $ tag =<< StringF . toS <$> escapedString
```
kim:
```suggestion
stringLiteralP = lexeme $ tag =<< StringF . toS <$> banana
  where
    banana = catMaybes <$> (char '"' >> manyTill ch (char '"'))
```

I think it should be called `banana`

PR #162, opened 2018-10-23
Update radicle files to reflect new eval

jkarni:
Some radicle files (in particular `chains.rad`) had become outdated after the new eval change. This PR updates them to work. Notable changes:

- The problem where defs wouldn't work after leaving a chain is now fixed
- `enter-chain` and `enter-remote-chain` are now special forms defined by an eval redefinition. They now have the form `(:enter-chain <chain>)` or `(:enter-remote-chain <chain>)`. 
- `:send` does not imply `:quit`

Changes that I also made that aren't directly related to entering chains:

- `map` now works with vectors
- Added `delete` primfn for dicts
- Added `rekey` function
- Show failure details in `should-be`

Still todo:

- [x] Clean up linger todos in `chain.rad`
- [x] `update-chain`
- [x] Finish up relevant docs section

### DIFF
--- a/rad/chain.rad
+++ b/rad/chain.rad
@@ -5,6 +5,11 @@
 ;;     == base-eval')
 ;;   - A sequence of inputs.
 
+
+;; DELETEME
+(def eval-definer (fn [ev] (write-ref eval ev)))
+;; DELETEME
+
 (def empty-chain
   (dict :state (pure-env)
         :inputs (list)
@@ -12,7 +17,7 @@
 
 (def eval-in-chain
   (fn [expr chain]
-    (def x (eval-with-env expr (lookup :state chain)))
+    (def x (eval expr (lookup :state chain)))
     (dict :chain (dict :state (head (tail x))
                        :input (cons expr (lookup :logs chain))
                        :index (+ 1 (lookup :index chain)))
@@ -21,6 +26,11 @@
 (document 'eval-in-chain
           '(("expr" any) ("chain" chain))
           "Evaluates 'expr' in the 'chain' and returns a dict with the ':result' and the resulting ':chain'.")
+
+(do 
+  (def res (eval-in-chain (+ 3 2) empty-chain))
+  (def ~~> (fn [comp expected] (should-be "eval-in-chain" (lookup comp) expected)))
+  (~~> :result 5))
 
 (def update-chain-with
   (fn [chain remote-chain-id cb-expr cb-res]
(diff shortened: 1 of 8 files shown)


rad/chain.rad
```diff
@@ -22,6 +30,15 @@
           '(("expr" any) ("chain" chain))
           "Evaluates 'expr' in the 'chain' and returns a dict with the ':result' and the resulting ':chain'.")
 
+(do 
+  (def res (eval-in-chain '(+ 3 2) empty-chain))
```
jameshaydon: These `def`s are global.

rad/chain.rad
```diff
@@ -38,52 +55,91 @@
...
+    (update-chain-with chain (lookup :name chain) (fn [x] (modify-ref next (fn [r] (eval-in-chain x r)))) print!)))
+
+(document 'update-chain
+		  '(("chain" chain))
+		  "Return a version of 'chain' with new inputs and correspondingly updated state")
+
+;; add-quit
+
+(def add-quit
+  (fn [old-state new-eval]
```
jameshaydon: Feels to me that the argument `new-eval` is more like `old-eval` (this is an eval-transformer function?).

rad/chain.rad
```diff
@@ -38,52 +55,91 @@
...
+                  (list :ok env_))
+              ((store-exprs oeval) expr env)))))
+
+(document 'add-send
+          '(("eval-fn" fn))
+          "Add a :send special form that sends the contents of _input to the chain _cur-chain")
+
+;; enter-chain & enter-remote-chain
+
+(def eval__ eval)
```
jameshaydon: Is this file part of the prelude? Or is this meant to capture "the current eval" when the file is `load!`ed?

rad/chain.rad
jkarni: Yeah, I think both 'old-state' and 'new-eval' are confusing names. I used them because I think of *entering* a new eval with the `:quit`, and then `:quit`-ing is going *back* to the old state. But I'll rename (and add some docs).

rad/chain.rad
jkarni: Yeah, so this is worth thinking a bit about. Currently it captures whatever eval happens to be in place at the moment this is run. If the only eval redefs we make are like this - in particular, if they're only adding commands, and commute, then I suppose it doesn't matter. But they won't always...

rad/chain.rad
jkarni: Switched to `fn`

rad/chain.rad
jameshaydon: I think it's okay for now.

PR #167, opened 2018-10-25
Do not use recursive eval for chain exploration

geigerzaehler:
Do not use recursive eval for chain exploration

### DIFF
--- a/rad/chain.rad
+++ b/rad/chain.rad
@@ -97,29 +97,27 @@
 ;; enter-chain & enter-remote-chain
 
 (def eval__ eval)
-(def-rec eval
+(def eval
   (fn [expr env]
       ;; The horror, the horror
       (if (list? expr)
           (if (> (length expr) 0)
               (cond
                 (eq? (head expr) :enter-chain) (do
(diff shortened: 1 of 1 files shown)


rad/chain.rad
```diff
@@ -97,23 +97,22 @@
...
                                                  (def mod-state
                                                    (over (.. (@ :env) (@ 'eval))
                                                          (fn [x] (add-quit env x))
-                                                         (lookup :state arg)))
+                                                         (lookup :state chain)))
                                                  (list :ok mod-state))
                 (eq? (head expr) :enter-remote-chain) (do
                                                         (def arg_ (nth 1 expr))
-                                                        (def arg (head (eval arg_ env)))
+                                                        (def chain (head (eval__ (nth 1 expr) env)))
```
jkarni: Any reason not to use `arg_` instead of `(nth 1 expr)`? (Or to not remove `arg_` as above)?

geigerzaehler: No. I fixed it

geigerzaehler:
This is not relevant anymore

PR #168, opened 2018-10-25
Get GHCJS client compiling again.

jkarni:


exe/Server/GHCJS.hs
```diff
@@ -62,48 +62,48 @@ jsEval bndsRef v = runInputT defaultSettings $ do
...
+          [String _, _] -> throwErrorHere $ TypeError "receive!: expecting number as second arg"
+          [_, _]        -> throwErrorHere $ TypeError "receive!: expecting string as first arg"
+          xs            -> throwErrorHere $ WrongNumberOfArgs "receive!" 2 (length xs)
       )
 
 -- * Helpers
 
 identV :: Text -> Value
-identV = Keyword . Ident
+identV = Keyword . unsafeToIdent
```
jameshaydon: We have a helper (or two) which does this already:
https://github.com/oscoin/radicle/blob/master/src/Radicle/Internal/PrimFns.hs#L324
and I think there is one in `Test` too.

jkarni: Yeah, but that's not exported since it's unsafe...

PR #170, opened 2018-10-26
Server persistence

### LINKED ISSUE #157: Server persistence
Add a persistence layer to `radicle-server` so data isn't all lost on crashes/restarts.

*Not* part of this task is a migration story for `radicle` updates that affect syntax or semantics.

jkarni:
Fixes #157.

### DIFF
--- a/exe/Server/Server.hs
+++ b/exe/Server/Server.hs
@@ -7,6 +7,7 @@ import qualified Data.Aeson as A
 import           Data.ByteString.Lazy (fromStrict)
 import qualified Data.Map as Map
 import qualified Data.Sequence as Seq
+import           Database.PostgreSQL.Simple (Connection, connectPostgreSQL)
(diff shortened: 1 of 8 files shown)


exe/Server/Server.hs
```diff
@@ -1,101 +1,203 @@
...
+    createIfNotExists conn
+    res <- getAllDB conn
+    chainPairs <- forM res $ \(name, vals) -> do
+        let st = runIdentity $ runLang pureEnv $ foldM_ (\_ x -> void $ eval x) () vals
+        case st of
+            (Left err, _) -> panic $ show err
+            (Right _, st') -> do
+                let c = Chain { chainName = name
+                              , chainState = st'
+                              , chainEvalPairs = mempty
```
jkarni: Hm, screwed up here. Need to keep the value pairs too

PR #171, opened 2018-10-26
Improve remote chain interaction

geigerzaehler:
This commit makes a couple of changes to improve the experience of interacting with remote chains.

It is best to start review in the documentation changes.

* Chains are identified by URLs instead of names. The `send!` and `receive!` operations now take a URL as the first parameter instead of a name.
* The `--url` option was removed. Since chains are identified by URLs there is no need for the `--url` parameter.
* `send!` does not include the chain name in the HTTP payload. Since chains are identified by URLs we now have a dedicated `/submit` endpoint for every chain and can just send the expression without the chain name.
* We removed all references to local chains from the docs and the Radicle code. We gear the experience towards remote chains which are more interesting.
* We only define the chain API endpoints in `exe/Server/API.hs`. All other endpoints exposed by the server are not important to `radicle-client`.
* If a chain does not exist its `/since` endpoint returns an empty list of Radicle values instead of a 400 status code. This is requied because we may load a non-existing chain with `:enter-chain`.

### DIFF
--- a/docs/source/guide/Chains.lrad
+++ b/docs/source/guide/Chains.lrad
@@ -31,51 +31,55 @@ In other words, we would like for programs to be *deterministic*.
 
 TODO: more.
 
-## Chains
+## Exploring remote chains
 
-### Remote chains
-
-The basic building blocks of interacting with remote chains are the `send!` and
-`receive!` operations. These send expressions to a remote chain, or receive new
-ones since a given index, respectively.
-
-These functions are, however, relatively low-level. Higher-level ones exist
-that do the grunt-work of only requesting data you don't already have, and of
-allowing interactive explorations of existing chains before submitting
-expressions. These are described in the next section.
-
-### Exploring chains
-
-Chains are a lot like shared environments.  You can enter an environment that
-corresponds to a chain and explore it:
+First we start up a simple server that hosts remote chains on port 8000.
 
 ```
-(def x 0)
-x                          ;; ==> 0
-(:enter-chain empty-chain) ;; ==> :ok
-(def x 1)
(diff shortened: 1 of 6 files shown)


geigerzaehler:
Note that the code is untested. We should figure out how to test this properly.

docs/source/guide/Chains.lrad
```diff
@@ -31,51 +31,55 @@ In other words, we would like for programs to be *deterministic*.
...
+## Exploring remote chains
 
-### Remote chains
+First we start up a simple server that hosts remote chains on port 8023.
 
-The basic building blocks of interacting with remote chains are the `send!` and
-`receive!` operations. These send expressions to a remote chain, or receive new
-ones since a given index, respectively.
+```
+stack exec -- radicle-server 8023
```
jkarni: This has changed in master in a couple of ways. First, we now need psql. Second, the port is a named arg.

exe/Server/Server.hs
```diff
@@ -61,6 +66,16 @@ data Chain = Chain
 -- matter for a test server.
 newtype Chains = Chains { getChains :: MVar (Map Text Chain) }
 
+type ServerApi
+    = "chains" :> Capture "chain" Text :> ChainSubmitEndpoint
+ :<|> "chains" :> Capture "chain" Text :> ChainSinceEndpoint
```
jkarni: I'd abstract the common path, since if they're *not* common everything breaks. I.e.:

```
"chains" :> Capture "chain" Text :> (ChainSubmitEndpoint :<|> ChainSinceEndpoint)
```

PR #173, opened 2018-10-29
Improve XDG files

geigerzaehler:
* Fix path for history file.
* Switch to more explicit `radicle` directory.
* Improve naming by appending `file` to exports.
* Improve documentation.

### DIFF
--- a/src/Radicle/Internal/CLI.hs
+++ b/src/Radicle/Internal/CLI.hs
@@ -7,7 +7,6 @@ where
 import           Protolude
 import           System.Directory (XdgDirectory(..), getXdgDirectory)
 
--- | Uses XDG_CONFIG_HOME if available.
 -- | Location of the radicle file to interpret.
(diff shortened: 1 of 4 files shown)


src/Radicle/Internal/CLI.hs
```diff
@@ -1,16 +1,25 @@
...
     )
 where
 
 import           Protolude
 import           System.Directory (XdgDirectory(..), getXdgDirectory)
 
 -- | Uses XDG_CONFIG_HOME if available.
-getConfig :: IO FilePath
-getConfig = getXdgDirectory XdgConfig "rad/config.rad"
+-- | Location of the default radicle file to interpretradicle history.
```
jkarni: I think the line above this should be deleted, or the pipe in this one.

Also, this isn't really related to history?
```suggestion
-- | Location of the radicle file to interpret.
```

PR #177, opened 2018-10-30
[WIP/RFC] Documentation functionality

jameshaydon:
This PR adds:
- A documentation type, which renders to `Pandoc` (and hence to most markup formats).
- Documentation data to environments. Each variable is optionally documented.
- Utilities to document primitives.
- A documentation-object value type.
- A 3-argument version of the `def` special form which takes a documentation object.
- A `doc!` primfn which prints the docs (as markdown).
- An `apropos!` primfn which prints docs for all variables in scope.
- Some primfns for creating documentation.

It's going to take a while to document all the primfns and all the prelude with this system so let's do some feedback before that, and maybe even merge too avoid conflicts.

Here are some examples:

Documenting a primfn:

```haskell
twoArg
  "eq?"
  [md|Checks if two values are equal.|]
  "x" [md|First value.|]  D.Any
  "y" [md|Second value.|] D.Any
  [md|Indicates if `x = y`.|]
  D.Boolean $
  \(a,b) -> pure $ Boolean (a == b)
```

And then in radicle REPL:
```
rad> (doc! 'eq?)
```

Produces:

---

`eq?`

Checks if two values are equal.

A `function`.

-   **Parameters:**

    `x`  
    First value. A `any`.

    `y`  
    Second value. A `any`.

-   **Returns:** Indicates if `x = y`. A `boolean`.

---

And in radicle one can use `def` like so:

```clojure
(def foo
  [:function
   "Adds 1 to numbers."
   [[x [:number "Some number."]]]
   [:number "The number `x` incremented by 1."]]
  (fn [x] (+ x 1)))
```

Which produces:

---

`foo`

Adds 1 to numbers.

A `function`.

-   **Parameters:**

    `x`  
    Some number. A `number`.

-   **Returns:** The number `x` incremented by 1. A `number`.

---

Docs for the prelude can be created in radicle by doing something like:

```clojure
(def prelude-docs
  (fn []
    (string-append
     "# Radicle Prelude\n\n"
     "## Primitive functions\n\n"
     (doc-md 'eq?)
     "\n\n"
     (doc-md 'uuid?))))
```


src/Radicle/Internal/Doc.hs
```diff
@@ -0,0 +1,191 @@
...
+
+instance Serialise (Described a) where
+  encode = notImplemented
+  decode = notImplemented
+
+instance ToPan a => ToPan (Described a) where
+  pan (Desc (Just d) x) = d <> pan x
+  pan (Desc _ x)        = pan x
+
+data Value
```
jkarni: Probably better named "type"?

src/Radicle/Internal/Doc.hs
```diff
@@ -0,0 +1,191 @@
...
+instance ToPan a => ToPan (Maybe a) where
+  pan Nothing  = Inlines []
+  pan (Just x) = pan x
+
+data Described a = Desc (Maybe Pan) a
+  deriving (Eq, Ord, Generic, Show, Read)
+
+instance Serialise (Described a) where
+  encode = notImplemented
+  decode = notImplemented
```
jkarni: umm

src/Radicle/Internal/Doc.hs
```diff
@@ -0,0 +1,191 @@
...
+      String -> typ "string"
+      Number -> typ "number"
+      Boolean -> typ "boolean"
+      Atom -> typ "atom"
+      Doc -> typ "doc"
+      Any -> typ "any"
+      Other -> Inlines []
+    where
+
+typ :: Text -> Pan
```
jkarni: I think you misformated this `where`

src/Radicle/Internal/Doc.hs
```diff
@@ -0,0 +1,191 @@
...
+blocks :: ToPan a => a -> Blocks
+blocks x = case pan x of
+  Blocks bs  -> bs
+  Inlines is -> [Para is]
+
+instance ToPan Blocks where
+  pan = Blocks
+
+instance ToPan a => ToPan (Maybe a) where
+  pan Nothing  = Inlines []
```
jkarni: This seems a little arbitrary. Why not `Blocks []` for instance?

(Is this class all that useful/sensible, rather than having multiple functions?)

src/Radicle/Internal/Effects.hs
```diff
@@ -79,17 +81,41 @@ replBindings = e { bindingsPrimFns = bindingsPrimFns e <> replPrimFns
...
 replPrimFns :: forall m. ReplM m => PrimFns m
-replPrimFns = PrimFns . Map.fromList $ first unsafeToIdent <$>
+replPrimFns = fromList $ Doc.noDocs $ first unsafeToIdent <$>
     [ ("print!", \case
         [x] -> do
             putStrS (renderPrettyDef x)
             pure nil
         xs  -> throwErrorHere $ WrongNumberOfArgs "print!" 1 (length xs))
 
+    , ( "doc!"
```
jkarni: This isn't actually impure, right? Is the idea that documentation raises questions about referential transparency?

src/Radicle/Internal/Pretty.hs
```diff
@@ -49,6 +49,7 @@ instance (Copointed t, Ann.Annotation t) => Pretty (Ann.Annotated t ValueF) wher
                         [ brackets . sep $ pretty <$> ids
                         , sep $ pretty <$> toList vals
                         ])
+        Doc _ -> angles "doc" -- TODO
```
jkarni: ...

src/Radicle/Internal/PrimFns.hs
```diff
@@ -339,13 +377,56 @@ purePrimFns = PrimFns $ fromList $ first Ident <$>
...
+oneArg :: Monad m =>
+          Text
+       -> D.Pan
+       -> Text
+       -> D.Pan
+       -> D.Value
+       -> D.Pan
+       -> D.Value
+       -> (Value -> Lang m Value)
+       -> (Ident, Maybe (D.Described D.Value), [Value] -> Lang m Value)
```
jkarni: This isn't the most wonderful function in the world. This arg stuff was already begging for reconsideration, but now I think it's been pushed over some edge. There are a few improvements to consider:

- a) Simplest: add a record datatype for one/two/threeArgs
- b) Add a single record type with list of args, no check that docs match number of args in fn
- c) Like (b), but use an hlist to make sure numbers do match
- d) Add singletons for prisms into `Value`, and make those singletons instance of a class that gives them types. Possibly make a class to recurse on number of args. Make sure number of arg documentation matches number of args.

src/Radicle/Internal/Effects.hs
jameshaydon: Well, it's as impure as `print!`.

jameshaydon: (It print's a formatted text to output, rather than return a string.)

src/Radicle/Internal/Pretty.hs
jameshaydon: "WIP"

src/Radicle/Internal/Doc.hs
jameshaydon: I think "type" is a bit confusing, it's documentation about a value.

src/Radicle/Internal/Doc.hs
jameshaydon: `Inlines []` makes it so that when combined with a piece of inline documentation, it doesn't promote to a block.

src/Radicle/Internal/PrimFns.hs
jameshaydon: What option do you prefer?

src/Radicle/Internal/Doc.hs
jameshaydon: Because I'm still not sure I want to have docs in values, so might not be necessary.

src/Radicle/Internal/Doc.hs
jameshaydon: Lol it's actually also used in `instance ToPan Function`.

jameshaydon:
@jkarni Replaced doc-objects with simple strings.

src/Radicle/Internal/PrimFns.hs
```diff
@@ -88,59 +101,73 @@ purePrimFns = PrimFns $ fromList $ first Ident <$>
...
+      , [md|Turns the arguments into a list.|]
+      , pure . List)
+    , ("dict"
+      , [md|Given an even number `2n` of arguments, creates a dict where the `2i`-th argument
+           is the key for the `2i+1`th argument.|]
+      , (Dict . foldr (uncurry Map.insert) mempty <$>)
                         . evenArgs "dict")
-    , ("throw", \case
+    , ("throw"
+      , [md|Throws an error. The first argument should be an atom used as a label for
```
jkarni: I'd say `exception` rather than `error`

src/Radicle/Internal/PrimFns.hs
```diff
@@ -88,59 +101,73 @@ purePrimFns = PrimFns $ fromList $ first Ident <$>
...
+           the exception, the second should be a string.|]
```
jkarni: The second can be any value.

src/Radicle/Internal/PrimFns.hs
```diff
@@ -88,59 +101,73 @@ purePrimFns = PrimFns $ fromList $ first Ident <$>
...
-    , ("cons", \case
+    , ("cons"
+      , [md|Adds an element to the front of a list.|]
+      , \case
           [x, List xs] -> pure $ List (x:xs)
           [_, _]       -> throwErrorHere $ TypeError "cons: second argument must be list"
           xs           -> throwErrorHere $ WrongNumberOfArgs "cons" 2 (length xs))
-    , ("head", oneArg "head" $ \case
+    , ("head"
+      , [md|Retreives the first element of a list if it exists. Otherwise throws an
```
```suggestion
      , [md|Retrieves the first element of a list if it exists. Otherwise throws an
```

src/Radicle/Internal/PrimFns.hs
```diff
@@ -43,13 +44,17 @@ addPrimFns primFns bindings =
              }
 
   where
-    primFnsEnv = Env $ Map.mapWithKey (\k _ -> PrimFn k) (getPrimFns primFns)
+    primFnsEnv = --Env $ Map.mapWithKey (\k _ -> PrimFn k) (getPrimFns primFns)
```
```suggestion
    primFnsEnv = Env (Map.fromList [ (pfn, Doc.Docd d (PrimFn pfn)) | (pfn, Doc.Docd d _) <- Map.toList (getPrimFns primFns)])
```

src/Radicle/Internal/PrimFns.hs
```diff
@@ -43,13 +44,17 @@ addPrimFns primFns bindings =
...
+      Env (Map.fromList [ (pfn, Doc.Docd d (PrimFn pfn)) | (pfn, Doc.Docd d _) <- Map.toList (getPrimFns primFns)])
```
```suggestion
```

src/Radicle/Internal/Effects.hs
```diff
@@ -73,26 +77,59 @@ replBindings :: forall m. ReplM m => Bindings (PrimFns m)
...
+      , [md|Reads a single line of input and returns it as a string.|]
+      , \case
+          [] -> String <$> getLineS
+          xs -> throwErrorHere $ WrongNumberOfArgs "get-line!" 0 (length xs)
+      )
 
-    , ("subscribe-to!", \case
+    , ("subscribe-to!"
+      , [md|Expects a dict `s` (representing a subscription) and a function `f`. The dict
+           `s` should have a function `getter` at the key `:getter`. This function iscalled
```
jkarni:
```suggestion
           `s` should have a function `getter` at the key `:getter`. This function is called
```

src/Radicle/Internal/Effects.hs
```diff
@@ -73,26 +77,59 @@ replBindings :: forall m. ReplM m => Bindings (PrimFns m)
...
+           repetitively (with no arguments), its result is then evaluated and passed to `f`.|]
```
jkarni:
```suggestion
           repeatedly (with no arguments), its result is then evaluated and passed to `f`.|]
```

jkarni:
@jameshaydon can you update the PR description to reflect the new design?

PR #184, opened 2018-11-01
Clearer radicleisation of documented environment variables.

jameshaydon:


jkarni:
Can you rebase against master so rad tests run?

jameshaydon:
But _of course_ they will fail ... :)

PR #188, opened 2018-11-02
F/validation

jameshaydon:
A very minimal validation lib (+ some related primitives).

Can be used as follows. Define a validator:

```clojure
(def val/issue
  (val/keys
   {:id     (val/and [(val/type :string)
                      (val/pred "valid UUID"
                                uuid?)])
    :author (val/pred "valid public key"
                      public-key?)
    :title  (val/and [(val/type :string)
                      (val/pred "< 60 chars"
                                (fn [s] (< (string-length s) 60)))])
    :body   (val/and [(val/type :string)
                      (val/pred "valid markdown"
                                markdown?)
                      (val/pred "< 4000 chars"
                                (fn [s] (< (string-length s) 4000)))])
    :tags   (val/all (val/and [(val/type :string)
                               (val/pred "< 40 chars"
                                         (fn [s] (< (string-length s) 40)))]))
    :priority (val/and [(val/type :number)
                        (val/pred "is integer"
                                  integral?)
                        (val/pred "0 <= p <= 10"
                                  (fn [p] (and (< -1 p) (< p 11))))])}))
```

Validate some data:

```
rad> (val/issue test-issue)
Exception validation-failure "Missing required key: :priority"
```

### DIFF
--- a/exe/ReferenceDoc.hs
+++ b/exe/ReferenceDoc.hs
@@ -16,8 +16,7 @@ import           Radicle.Internal.Doc (md)
 import           System.Console.Haskeline (defaultSettings, runInputT)
 import           Text.Pandoc
 
-main ::
-  IO ()
+main :: IO ()
 main = do
     res_ <- runInputT defaultSettings $
              interpret
@@ -62,6 +61,10 @@ main = do
             "Effectful functions. These functions are not available in 'pure' chains, but are available in the local REPL."
       , sec "Lenses" lens
             "Functional references into radicle values."
+      , sec "Validation" validation
+            [md|Functions for creating or combining *validators*, which are functions which return the
+               input unchanged or throw with an error message. These can be used for checking data before
+               accepting it onto a chain.|]
       , sec "Cryptography" crypto
             "Tools for creating and verifying cryptographic signatures, and generating private/public key pairs."
       , sec "Chain tools" chainTools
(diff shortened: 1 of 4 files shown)


rad/prelude/validation.rad
```diff
@@ -0,0 +1,57 @@
+(def val/=
+  "Given `x`, returns a validator that checks for equality with `x`."
+  (fn [x]
+    (fn [y]
+      (if (eq? x y)
+        x
+        (throw 'validation-failure (string-append "Not equal to: " (show x)))))))
+
+(def val/and
```
jkarni: Maybe?
```suggestion
(def val/all
```

rad/prelude/validation.rad
```diff
@@ -0,0 +1,57 @@
...
+      (if (dict? d)
+        ((val/and
+          (map (fn [kv] (val/key (nth 0 kv) (nth 1 kv)))
+               (seq ks)))
+         d)
+        (throw 'validation-failure (string-append "Not a dict: " (show d)))))))
+
+(def val/all
+  "Check that all the items of a sequence match a validator."
+  (fn [v]
```
jkarni: Hm, confusing what the distinction is to `and`

jameshaydon: shall we stick with `val/and` and rename `val/all` to `val/every`?

jameshaydon: @jkarni Shall we keep `val/and` and rename `val/all` to `val/every`?

src/Radicle/Internal/PrimFns.hs
```diff
@@ -166,6 +169,18 @@ purePrimFns = fromList $ allDocs $
           _           -> throwErrorHere $ TypeError "tail: expects list argument")
 
     -- Lists and Vecs
+    , ( "drop"
```
jkarni: Some tests for these?

src/Radicle/Internal/PrimFns.hs
```diff
@@ -166,6 +169,18 @@ purePrimFns = fromList $ allDocs $
...
+      , [md|Returns all but the first item of a sequence, unless the sequence is empty,
```
```suggestion
      , [md|Returns all but the first `n` items of a sequence, unless the sequence is empty,
```

exe/ReferenceDoc.hs
```diff
@@ -105,7 +110,7 @@ main = do
 
     allFns =
          basics ++ maths ++ evalFns ++ envStuff ++ seqs ++ lists ++ vecs
-      ++ dicts ++ structs ++ refs ++ docs ++ io ++ lens ++ crypto ++ chainTools
+      ++ dicts ++ structs ++ refs ++ docs ++ io ++ lens ++ validation ++ crypto ++ chainTools
```
jkarni: Maintaining this is very quickly going to get annoying. Should it perhaps not be automatically generated from the file order, with perhaps some way of documenting a module intro (or sections)?

jkarni: I take this back. The tooling around this is already super nice! Hats off!

PR #192, opened 2018-11-02
Fixes #191 in which catch did not correctly handle the scope of the handler

### LINKED ISSUE #191: Scoping is broken in (catch)
The following test:

```
(should-be "catch handles scope correctly"
  (do 
    (def catch-f (fn [l body handler]
      (catch l (body) handler)))
    (catch-f 'any (fn [] (throw 'foo "foo")) (fn [e] #t)))
  #t)
```

Fails with `Unknown identifier: handler`

It seems that `catch` isn't correctly handling scope in its handler argument.

luqui:


luqui:
Note that this makes `catch`'s evaluation order somewhat weird.  Is there another way?

PR #201, opened 2018-11-07
Improve chain API

geigerzaehler:
We stop using the`Show` and `Read` instances to serialise Radicle values for the plaintext MIME type. Instead we render the radicle values as code and parse them as code.

This also changes the contract of the `since` endpoint. Now we return the expresssions as a list of Radicle values.

### DIFF
--- a/exe/API.hs
+++ b/exe/API.hs
@@ -7,19 +7,24 @@ import           Protolude
 import           Radicle
 import           Servant.API
 
-instance Show a => MimeRender PlainText a where
-    mimeRender _ x = show x
+instance MimeRender PlainText Value where
+    mimeRender _ x = LBS.fromStrict $ encodeUtf8 $ renderCompactPretty x
 
-instance Read a => MimeUnrender PlainText a where
-    mimeUnrender _ x = readEither $ T.unpack $ decodeUtf8 $ LBS.toStrict x
+instance MimeUnrender PlainText Value where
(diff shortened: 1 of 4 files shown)


exe/API.hs
```diff
@@ -7,19 +7,24 @@ import           Protolude
...
+instance MimeUnrender PlainText Value where
+    mimeUnrender _ x =
+        let src = decodeUtf8 $ LBS.toStrict x
+        in first T.unpack $ parse "[request body]" src
 
-
-type ChainSubmitEndpoint = "submit" :> ReqBody '[PlainText] Value :> Post '[PlainText] ()
+-- | Endpoint to submit an expression to a chain. Responds with @:ok@ if the
+-- expression has been submitted sucessfully.
+type ChainSubmitEndpoint = "submit" :> ReqBody '[PlainText] Value :> Post '[PlainText] Value
```
jkarni: If the success response is always just `:ok`, why have it at all?

geigerzaehler: And alternative to this would be to have `()` as the return type for the endpoint. This would require the orphan instance `MimeRender PlainText ()` which I don’t like. I also think it’s nice to have a consistent API where the response is always a Radicle value. Especially if we want to move to a Radicle MIME type.

jkarni: :+1:

PR #203, opened 2018-11-08
Add notes for Radicle P2P on hypercore

geigerzaehler:
This adds a document that provides a rough overview of how Radicle P2P can work on top of `hypercore`. (`hypercore` is a library that is part of the Dat stack.)

[Preview](https://github.com/oscoin/radicle/blob/radicle-p2p-notes/notes/radicle-p2p-hypercore-overview.md)

*Todo*
- [ ] Addressing of entries in a multi-writer scenario
- [ ] Properties of simple multi-writer consensus algorithm
- [ ] Add details on replication and offline use
- [ ] Explain how we would implement this

### DIFF
--- a/notes/radicle-p2p-hypercore-overview.md
+++ b/notes/radicle-p2p-hypercore-overview.md
@@ -1,0 +1,133 @@
+Radicle Peer-to-peer Using `hypercore`
+======================================
+
+This document provides a high-level overview of how to build a P2P peer system
+to distribute and collaborate on Radicle chains using the
+[`hypercore`][hypercore] and [`discovery-swarm`][discovery-swarm] libraries.
+
+[hypercore]: https://github.com/mafintosh/hypercore
+[discovery-swarm]: https://github.com/mafintosh/discovery-swarm
+
+### Status of This Document
+
+This document is only for informational purposes and to summarize our research
+into `hypercore`. We decided not to build Radicle P2P on `hypercore`. You can
+see the discussion about this document in
+[PR#203](https://github.com/oscoin/radicle/pull/203).
+
+
+Hypercore
+---------
+
+`hypercore` is a library to distribute cryptographically secure append-only logs
+called *feeds*. A feed consists of a list of *entries* which may be JSON objects
+or binary blobs. A feed is signed with a private key. The hash of the
+corresponding public key is called the *discovery key* and acts as an identifier
+for the feed.
+
+Any user who knows the location (for example the IP and port) of a feed can read
+any entry in the feed, verify the integrity of any entry, listen to new entries,
+and replicate the feed.
+
+We say a user *hosts a feed* if the user makes it available to the world. If a
+users hosts a feed the user is either the owner of that feed or she replicates
+it.
+
+`hypercore` supports feed subscriptions. A user can subscribe to a feed and will
+be notified if the owner publishes a new entry.
+
+Every feed a user owns or replicates is stored locally and updated automatically
+if a user goes online.
+
+`discovery-swarm` is a library to discover hosts for hypercore feeds and
+announce the hosting of a feed. It uses the Bittorrent DHT and centralised DNS
+for global discovery as well as Multicast DNS for local discovery.
+
+
+Single-writer Chains
+--------------------
+
+*Single-writer chains* are Radicle chains that only the owner of a specific
+private key can write to.
+
+We use hypercore feeds to represent single-writer chains. Each entry in the feed
+is a Radicle expression. Single-writer chains are identified by the discovery
+key of their feed.
+
+To read a Radicle chain a user needs the discovery key of a feed. With the
+discovery key the user uses hyperdiscovery to find someone who replicates the
+feed. (This may be the owner.) The user then reads all entries in the feed with
+hypercore and evaluates them with radicle.
+
+To write an expression to a Radicle chain a user publishes a new entry to the
+`hypercore` feed.
+
+
+Multi-writer Chains
+-------------------
+
+*Multi-writer chains* are Radicle chains that the owners of multiple distinct
+private keys can write to.
+
+We implement a multi-writer chain by using set of feeds owned by different
+users. Every entry in these feeds is a Radicle expression and a link to another
+entry in any of feeds. Every entry in a feed now determines a linked list of
+Radicle expression (assuming the feed entries and links are well formed and
+valid) and thus a Radicle chain.
+
+To derive a Radicle chain from a set of feeds we need an algorithm that takes
+the heads of all the feeds of a multi-writer chain and outputs either a chain of
+Radicle expressions or signals that the chain is invalid.
+
+The simplest such algorithm is the following: We accept only the situation where
+there is a feed with head entry `e` so that the head entries of all other feeds
+are ancestors of `e`. In that case `e` is considered the head of the chain.
+Otherwise the chain is invalid.
+
+
+**TODO** Properties of this scheme
+
+**TODO** Addressing entries
+
+
+Offline First and Replication
+-----------------------------
+
+**TODO** flesh this out
+
+Hypercore stores every feed a user is interested locally.
+
+Hypercore allows one user to replicate the feed of another user. Updates are
+propagated automatically.
+
+Users can fetch a feed from a user replicating it if the owner is not online.
+
+Implementation
+--------------
+
+`hypercore` and `discovery-swarm` are only available as JavaScript libraries. To
+build Radicle P2P on top of hypercore we would need to implement the core
+functionality of a Radicle P2P node in JavaScript. (Typescript and Purescript
+are alternative languages since they can use JavaScript transparently.) This
+node would expose an API that would allow the Radicle REPL and other tools to
+submit and get expressions.
+
+The node will be accompanied by a control CLI that allows users to create
+chains, replicate chains, etc.
+
+Security
+--------
+
+* Feeds can be censored. If the owner goes offline and all users that replicate
+  the feed conspire to drop part of the head of the feed. If one user chooses to
+  replicate the actual head of the feed then everyone else will see that head,
+  too.
+* Feeds are unforgeable. Anybody with the public key of a feed can verify
+  whether a given entry was published to the feed by its owner.
+
+Open Questions
+--------------
+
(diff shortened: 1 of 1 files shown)


notes/radicle-p2p-hypercore-overview.md
```diff
@@ -0,0 +1,114 @@
...
+Radicle expression (assuming the feed entries and links are well formed and
+valid) and thus a Radicle chain.
+
+To derive a Radicle chain from a set of feeds we need an algorithm that takes
+the heads of all the feeds of a multi-writer chain and outputs either a chain of
+Radicle expressions or signals that the chain is invalid.
+
+The simplest such algorithm is the following: We accept only the situation where
+there is a feed with head entry `e` so that the head entries of all other feeds
+are ancestors of `e`. In that case `e` is the chain derived from the set of
```
jkarni:
```suggestion
are either `e` or ancestors of `e`. In that case `e` is the tip chain derived from the set of
```

geigerzaehler: > are either `e` or ancestors of `e`.

I don’t think there is a sensible notion of equality of entries in different feeds that would support this statement. Entries are addressed by their feed key and the sequence number. So by definition entries in different feeds will have different addresses. One could say that entries are the same if there content is the same. But this would require every collaborator to replay entries from every other collaborator to their chain.

> the tip chain

What does the term “tip chain” mean?

jkarni: I see your point about ancestors. I was thinking we were just porting over the IPFS idea, where it is possible for the multiple IPNSs to point to exactly the same hash, but I like your idea better.

> What does the term “tip chain” mean?

 I think `head` might be the term you're using. But as it stands, it seems wrong to me: `e` isn't the chain itself, it's just the latest entry in it.

geigerzaehler: > > What does the term “tip chain” mean?
> 
> I think `head` might be the term you're using. But as it stands, it seems wrong to me: `e` isn't the chain itself, it's just the latest entry in it.

Yeah, right. `e` refers to an entry. I’ll reformulate it.

notes/radicle-p2p-hypercore-overview.md
```diff
@@ -0,0 +1,114 @@
...
+  too.
+* Feeds are unforgeable. Anybody with the public key of a feed can verify
+  whether a given entry was published to the feed by its owner.
+
+Open Questions
+--------------
+
+* How robust and secure are `hypercore` and `discovery-swarm`?
+* How many feeds can a node host? From the examples and the code it looks
+  like a node can host one feed per port.
```
kim: > one feed per port

:eyes:

notes/radicle-p2p-hypercore-overview.md
```diff
@@ -0,0 +1,114 @@
...
+propagated automatically.
+
+Users can fetch a feed from a user replicating it if the owner is not online.
+
+
+Security
+--------
+
+* Feeds can be censored. If the owner goes offline and all users that replicate
+  the feed conspire to drop part of the head of the feed. If one user chooses to
```
kim: > drop part of the head

Drop the head? 

Is that an actual concern? Replicators can conspire to not propagate the latest update.

kim:
General question: how "deep" would we need to integrate? Ie. do we need to use JS APIs to do anything useful, or could we also just shell out to some Dat executable?

notes/radicle-p2p-hypercore-overview.md
geigerzaehler: > Is that an actual concern? Replicators can conspire to not propagate the latest update.

I don’t think it is an actual concern. I just wanted to make a note of it.

geigerzaehler:
> General question: how "deep" would we need to integrate? Ie. do we need to use JS APIs to do anything useful, or could we also just shell out to some Dat executable?

I added an “Implementation” section addressing this. TL;DR we need to build a JavaScript node.

notes/radicle-p2p-hypercore-overview.md
```diff
@@ -0,0 +1,126 @@
...
+chains, replicate chains, etc.
+
+Security
+--------
+
+* Feeds can be censored. If the owner goes offline and all users that replicate
+  the feed conspire to drop part of the head of the feed. If one user chooses to
+  replicate the actual head of the feed then everyone else will see that head,
+  too.
+* Feeds are unforgeable. Anybody with the public key of a feed can verify
```
jameshaydon: Can the owner forge the feed? E.g. can they change the first entry.

jameshaydon:
Some notes regarding the multi-writer chains:

> We implement a multi-writer chain by using set of feeds owned by different users.

How is this set defined/determined? What's stopping another feed claiming membership to this set and linking to previous entries?

> To derive a Radicle chain from a set of feeds we need an algorithm that takes the heads of all the feeds of a multi-writer chain and outputs either a chain of Radicle expressions or signals that the chain is invalid.

Since the entries form a DAG, any topological sorting might be considered valid.

jkarni:
> How is this set defined/determined? What's stopping another feed claiming membership to this set and linking to previous entries?

One thing we talked about is the first entry in a feed stating what all the (initial) writers are. That would also help with discovery.

notes/radicle-p2p-hypercore-overview.md
```diff
@@ -0,0 +1,126 @@
...
+
+
+Multi-writer Chains
+-------------------
+
+*Multi-writer chains* are Radicle chains that the owners of multiple distinct
+private keys can write to.
+
+We implement a multi-writer chain by using set of feeds owned by different
+users. Every entry in these feeds is a Radicle expression and a link to another
```
clevinson: What does it mean if Alice wants to reference one of Bob's entries in a feed of hers? Does she:
1.   Fork bob's feed ?
2.  Create an entry in her already existing feed that has 2 pointers (one to the previous entry in her own feed, and one to this entry in bob's feed that she wants to reference) ?

If 1: What is the hypercore "public key" of her fork of bob's feed in this sense? how does her fork get represented as hypercore feeds?
If 2: Does multiple pointers in a feed entry mean multiple states/environments that a radical expression references?

jameshaydon:
> Every entry in a feed now determines a linked list of Radicle expression (assuming the feed entries and links are well formed and valid) and thus a Radicle chain.

While it's true that every entry determines a chain (by going back up to the root), the general structure is a tree. How do we avoid branches?

notes/radicle-p2p-hypercore-overview.md
geigerzaehler: If the owner does anything but append an entry to the feed and a version of the previous state of the feed is still being replicated the feed is treated as corrupted.

Feeds provide the same security guarantees as hash linked lists. Changing any entry in the feed will change the overall hash of all following entries and consequently the hash of the current state of the feed.

notes/radicle-p2p-hypercore-overview.md
geigerzaehler: I’m not sure what you mean by “reference”. As I see it an entry in a feed only every link to one other entry in another feed. This is the link to the parent entry which determines the whole chain of Radicle expressions to follow.

notes/radicle-p2p-hypercore-overview.md
kim: > the feed is treated as corrupted

There isn't really a way to recover from that, is it?

hxrts:
> > How is this set defined/determined? What's stopping another feed claiming membership to this set and linking to previous entries?
> 
> One thing we talked about is the first entry in a feed stating what all the (initial) writers are. That would also help with discovery.

Don't we ultimately want multiwriter permissions to be modifiable via radicle eval updates in order to generalize the chain permissions example given in the radicle whitepaper?

geigerzaehler:
> Don't we ultimately want multiwriter permissions to be modifiable via radicle eval updates in order to generalize the chain permissions example given in the radicle whitepaper?

Yes, we want that. But this is impossible unless the storage system is able to understand radicle because it needs to authenticate it. I think the only chance we have to make this work is a block chain with validators evaluating radicle.

jkarni:
> Yes, we want that. But this is impossible unless the storage system is able to understand radicle because it needs to authenticate it. I think the only chance we have to make this work is a block chain with validators evaluating radicle.

Can you explain this a bit? I'd have thought that if we have the original expressions list the initial writers, it'd simply be a matter of any of those writers changing that original set, no? (Whether authentication happens at the storage layer or at the radicle layer.)

geigerzaehler:
> > Yes, we want that. But this is impossible unless the storage system is able to understand radicle because it needs to authenticate it. I think the only chance we have to make this work is a block chain with validators evaluating radicle.
> 
> Can you explain this a bit? I'd have thought that if we have the original expressions list the initial writers, it'd simply be a matter of any of those writers changing that original set, no? (Whether authentication happens at the storage layer or at the radicle layer.)

What you are suggesting unfortunately does not work. If the storage system does not know about Radicle it cannot decide whether to accept a Radicle expression based on the current state of the chain. Conversely consumers will be unable to verify that a user has permission to write to a chain without evaluating the Radicle expressions on the chain.

> By "blockchain" you don't mean a BFT system, just that everyone is running their own node to check if the signatures are valid before adding to the chain, correct?

Yes.

>  If a new chain is instantiated with an authorized set of writers, it seems like any of those writers should be able to change the key set. There are some interesting design choices around when original writers are removed, or whether we eventually want to allow forward encryption of messages, but I think we could standardize how those cases are handled, no?

This is an excellent point and we should definitely discuss. At the moment I don’t think we have enough about this, though. So we should postpone this discussion until an actual need arises.

jkarni:
> What you are suggesting unfortunately does not work. If the storage system does not know about Radicle it cannot decide whether to accept a Radicle expression based on the current state of the chain. Conversely consumers will be unable to verify that a user has permission to write to a chain without evaluating the Radicle expressions on the chain.

Hm, my mental model was precisely that: the storage system doesn't know what to accept - only people consuming a chain do (by running the chain). Isn't that kind of a) unavoidable, and b) completely fine?

geigerzaehler:
Since we decided to use IPFS after all I have added a paragraph that marks this document as purely informational. I would nevertheless like to merge it so we don’t loose it. Could somebody mark this as reviewed?

We can continue the open discussions within the scope of the IPFS proposal.

PR #211, opened 2018-11-13
Replace Scientific with Rational

jameshaydon:


src/Radicle/Internal/PrimFns.hs
```diff
@@ -269,9 +270,9 @@ purePrimFns = fromList $ allDocs $
     --
     -- In order to avoid this sort of thing, we don't allow +,*,- and / to be
     -- applied to a single argument.
-    , numBinop (+) "+" [md|Adds two numbers together.|]
-    , numBinop (*) "*" [md|Multiplies two numbers together.|]
-    , numBinop (-) "-" [md|Substracts one number from another.|]
+    , numBinop (Num.op (+)) "+" [md|Adds two numbers together.|]
+    , numBinop (Num.op (*)) "*" [md|Multiplies two numbers together.|]
+    , numBinop (Num.op (-)) "-" [md|Substracts one number from another.|]
```
jkarni: Can you add `/`? We can finally have it, and it'll establish more clearly the coercion rules/numeric tower.

jkarni: Hm, actually, having it will still be tricky, since division of scientifics is a mess...

PR #214, opened 2018-11-13
docs: Remove Pandoc dependency

xla:
As most of all inputs are valid markdown the existing checks didn't serve any real purpose, but made us depend on Pandoc. Which broke the ghcjs build.

### DIFF
--- a/docs/source/reference.rst
+++ b/docs/source/reference.rst
@@ -1,0 +1,883 @@
+Radicle Reference
+=================
+
+These are the functions that are available in a new radicle chain after
+the prelude has been loaded.
+
+Basics
+------
+
+Basic function used for checking equality, determining the type of a
+value, etc.
+
+``eq?``
+~~~~~~~
+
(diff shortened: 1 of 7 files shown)


jameshaydon:
There are some problems with the multiline strings. Sometimes there is a space missing between words, and sometimes there is a (uselss?) newline and some indentation.

jkarni:
I didn't notice any extra newlines while fixing, but I also didn't notice any in the generated ref docs. Anyhow, extra spaces should now be fixed.

jameshaydon:
> I also didn't notice any in the generated ref docs

The reference docs go through pandoc, so newlines would be cleaned up. But actually the newlines/indentations I saw (using `(apropos!)`) are in docstrings of prelude functions. Maybe we could tweak `def` a bit to clean this up.

jkarni:
Btw, including `reference.rst` since RTD can't build it automatically.

PR #217, opened 2018-11-14
Dict keys should be hashable.

### LINKED ISSUE #112: Consider dictionary semantics more carefully.
Radicle is meant to be a fully specified language, but the semantics of dictionaries are rather bizarre.

The default ordering of values is relevant in the evaluation of a dict (when resolving key-collisions), or just when calling `seq`, and _all_ values are permitted as keys in dicts. Thus for example:

```scheme
(define a (ref 0))
(define b (ref 0))
...
(define d {a 0 b 1})
(seq d)
```

The resulting list is ordered according to the _addresses_ in the refs, but the (order of the) addresses of refs should not enter in the semantics of the language. Things are even more bizarre when using functions as keys!

**Proposal:** Creating a dict with non-[hashable](https://github.com/oscoin/radicle/issues/106#issuecomment-424114048) keys results in an error.

jameshaydon:
Closes #112.

### DIFF
--- a/src/Radicle/Internal/Core.hs
+++ b/src/Radicle/Internal/Core.hs
@@ -152,13 +152,17 @@ instance Serialise r => Serialise (ValueF r)
 
 hashable :: (CPA t) => Annotated t ValueF -> Bool
 hashable = \case
+  Atom _ -> True
+  Boolean _ -> True
+  Number _ -> True
+  String _ -> True
+  Keyword _ -> True
   PrimFn _ -> False
   Ref _ -> False
   Lambda{} -> False
   List xs -> all hashable xs
   Vec xs -> all hashable xs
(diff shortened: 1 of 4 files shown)


src/Radicle/Internal/Core.hs
```diff
@@ -148,6 +150,23 @@ data ValueF r =
...
 
+hashable :: (CPA t) => Annotated t ValueF -> Bool
+hashable = \case
+  PrimFn _ -> False
+  Ref _ -> False
+  Lambda{} -> False
+  List xs -> all hashable xs
+  Vec xs -> all hashable xs
+  Dict kvs -> getAll $ Map.foldMapWithKey (\k v -> All (hashable k && hashable v)) kvs
+  _ -> True
```
geigerzaehler: Probably best to list all constructors explicitly. If a constructor is added we want the author to make an explicit decision here.

PR #229, opened 2018-11-19
Monadic issues chain

jameshaydon:


exe/ReferenceDoc.hs
```diff
@@ -69,8 +70,8 @@ main = do
             "These functions can be used to simulate remote chains in the local REPL.\
             \ This is useful for experimenting with inputs or even new evaluation functions\
             \ before sending these to a remote chain."
-      , sec "Issue chain" issueChain
-            "These functions allow creating and interacting with the default issues chain."
+      -- , sec "Issue chain" issueChain
```
jkarni: Delete?

exe/ReferenceDoc.hs
```diff
@@ -115,8 +119,9 @@ main = do
 
     allFns =
          basics ++ maths ++ evalFns ++ envStuff ++ seqs ++ lists ++ vecs
-      ++ dicts ++ structs ++ refs ++ docs ++ io ++ lens ++ validation ++ crypto ++ chainTools
-      ++ issueChain ++ doNotInclude
+      ++ dicts ++ sets ++ structs ++ refs ++ docs ++ io ++ lens ++ validation ++ crypto ++ chainTools
+      -- ++ issueChain
```
jkarni: delete?

rad/monadic/issue-remote.rad
```diff
@@ -0,0 +1,183 @@
...
+
+(def validator/issue-input
+  "An issue should have all the required fields, be a signed input, etc."
+  (validator/and
+   [validator/input
+    (validator/keys
+     {:title    (validator/text 100)
+      :body     (validator/text 4000)
+      :labels   (validator/every (validator/text 80))
+      :state    (validator/member [:open :closed])
```
jkarni: I keep being impressed with your validation stuff!

rad/monadic/issue-remote.rad
```diff
@@ -0,0 +1,183 @@
...
+      :comments (validator/every validator/comment)})]))
+
+(def strip-input
+  "After an entity input has been processed/verified, we no longer need some of
+  the fields."
+  (fn [e]
+    (delete-many [:chain-id :signature :nonce :issue-number] e)))
+
+(def create-issue
+  "Create an issue from a dict, checking it is valid and storing it in `issues`."
```
jkarni:
```suggestion
  "Create an issue from a dict, checking that it is valid and storing it in `issues`."
```

rad/monadic/issue-remote.rad
```diff
@@ -0,0 +1,183 @@
...
+  (lookup :next (make-counter)))
+
+(def used-nonces
+  "We keep track of the UUIDs which have been used as nonces in inputs."
+  (ref set/empty))
+
+(def unused-nonce?
+  "Check that a nonce has not been used already."
+  (fn [nonce]
+    (not (set/member? nonce (read-ref used-nonces)))))
```
jkarni: Btw, I think you can add tests to this file. Just use `:test`, and then in `Tests.hs` run a modified version of it (i.e., one that imports `prelude.rad`)

rad/monadic/issues.rad
```diff
@@ -0,0 +1,142 @@
...
+       (def pk (gen-key-pair! (default-ecc-curve)))
+       (def chain-name (uuid!))
+       (def chain-ref (create-issues-chain! chain-name))
+       (new-issue! chain-ref pk "Ah!" "if thou issueless shall hap to die")
+       (update-chain-ref chain-ref)
+       (def issues (values (list-issues chain-ref))))
+   ]
+   [ (length issues) ==> 1 ]
+   [ (lookup :title (nth 0 issues)) ==> "Ah!" ]
+   [ (lookup :body (nth 0 issues)) ==> "if thou issueless shall hap to die" ]
```
jkarni: Definitely needs more tests

rad/prelude/util.rad
```diff
@@ -0,0 +1,8 @@
+(def make-counter
+  "Creates a stateful counter. Returns a dict with two keys: the function at `:next-will-be`
+  will return the next number (without incrementing it), while the function at `:next`
+  increments the number and returns it."
+  (fn []
+    (def i (ref -1))
+    {:next-will-be (fn [] (+ 1 (read-ref i)))
```
jkarni: I think we generally should try to make things more private - here in theory someone could accidentally modify `next-will-be` or `next` and get things wrong. I'm imagining:

```
(def make-counter
   (fn []
     (def i (ref -1))
     (fn [command]
       (if (eq? command :next) (modify-ref ...)
           (if eq? command :current) ....)))
```

And then:

```
(def counter (make-counter))
(counter :current) ;; ==> 0
(counter :next) 
(counter :current) ;; ==> 1
```

What do you think?

rad/prelude/util.rad
```diff
@@ -0,0 +1,8 @@
...
+     :next         (fn [] (modify-ref i (fn [x] (+ x 1))))}))
```
jkarni: Also, test please!

rad/prelude/validation.rad
```diff
@@ -79,3 +79,24 @@
...
+  (validator/and
+   [(validator/keys
+     {:author (validator/pred "valid public key" public-key?)})
+    (validator/pred
+     "Valid signature"
+     (fn [d]
+       (verify-signature
+        (lookup :author d)
+        (lookup :signature d)
+        (show (delete :author (delete :signature d))))))]))
```
jkarni: Test please

src/Radicle/Internal/PrimFns.hs
```diff
@@ -239,6 +244,16 @@ purePrimFns = fromList $ allDocs $
             pure . Dict . Map.fromList $ zip (fst <$> kvs) vs
           _ -> throwErrorHere $ TypeError $ "map-values: second argument must be a dict"
       )
+
+    -- Structures
+    , ( "<>"
+      , "Merges two structures together. On vectors this performs concatenations. On dicts this performs the right-biased merge."
+      , twoArg "<>" $ \case
+          (Vec xs, Vec ys) -> pure $ Vec (xs Seq.>< ys)
```
jkarni: Might as well add lists?

rad/prelude/util.rad
jameshaydon: > someone could accidentally modify `next-will-be`

Not sure what you mean? They are immutable functions how could they be modified? In any case I think what you propose is a little clearer so will change.

rad/prelude/validation.rad
jameshaydon: I added a test but it doesn't work. I suspect the `catch` has something to do with it.

rad/monadic/issues.rad
jameshaydon: Added a test but it doesn't work, whenever I put more than one test it fails.

jkarni:
Closing in favor of #233

PR #230, opened 2018-11-19
Pattern matching

### LINKED ISSUE #228: Pattern matching
Somehow, despite talking about this quite a bit, we don't seem to have an issue for it.

We talked about unification-based systems earlier, and are still digging into Barry Jay's work. But the current thinking is some variation of Scala or Tullsen's "First Class Patterns". I.e., a pattern is just a function that either returns a maybe of values (that are then bound to whatever parameters are in the pattern) or a function that takes the thing being pattern matched and the pattern match itself, and constructs a dictionary of ident -> values.

jameshaydon:
Adds pattern matching with first-class patterns.

Pattern matching is invoked using `match`, here is a function with lots of matching:

```clojure
(def f
  (fn [x]
    (match x
      ;; constants
      1                                 :number-one
      ;; predicates
      (/? keyword?)                     [:kw x]
      ;; guards
      ((/? number?) | (> x 42))         [:bigger-than-42 x]
      (/? number?)                      :non-1-number
      ;; more constants
      :hello                            :hello
      "hola"                            "que tal"
      ;; A custom pattern
      (/Just 'x)                        x
      :Nothing                          :Nothing
      ;; Some listy stuff
      (/cons 1 (/cons 2 'xs))           [:one-two xs]
      (/cons 1 (/cons _ (/cons 3 'xs))) [:one-something-three xs]
      (/cons 'x (/cons 'y 'xs))         [:non-singleton x y xs]
      (/cons 'x _)                      [:singleton x]
      /nil                              :empty-list
      ;; Pattern matching on dicts:
      (/keys {:a 'a :b 'b})             [:a-was a :b-was b]
      ;; Even the wildcard is user-defined!
      _                                 [:anything-else x])))
```

Only `match` is a special form, and all it does is define evaluation order and syntax for guards. Everything else is defined in the prelude. This is possible because pattern matching calls out to a special dispatch function (for dealing with constants, variables and possibly other special patterns), and because non-special patterns are just functions. For example here is the pattern for `/Just`:

```clojure
(def /Just
  (fn [pat]
    (fn [v]
      (if (vector? v)
        (if (eq? (length v) 2)
          (if (eq? (nth 0 v) :Just)
            (match-pat pat (nth 1 v))
            :Nothing)
          :Nothing)
        :Nothing))))
```

Even the handling of special patterns could be modified if we stored the pattern dispatch function in a ref.

Closes #228.


jameshaydon:
Questions:

- Are we okay with quoted variables in patterns.
- Need as-patterns.
- Nested guards?

jkarni:
> Using an operation other than <> would make some of the patterns support non-linearity.

Non-linearity would be nice, but if we *don't* support it, we *should* throw an error on patterns which bind the same variable multiple times rather than picking one of those bindings.

jkarni:
This is really really cool! I very much like the basic desigs; the questions left over are relatively minor.

- Is there a reason to have `/keys` rather than just check that the value is a dict?
- Do guards need to be a *separate* thing? Conceptually, they feel like just a pattern that binds nothing (but can of course fail). So it feels like `|` can be generalized, making it in essence apply the pattern match twice. In the case where the rhs is a guard, we get back what we had.

rad/prelude/patterns.rad
```diff
@@ -0,0 +1,192 @@
...
+       [ (match {:a 1 :b [1 2 3]}
+                42 :no
+                ({:a _ :b [_ 'n _]} | (eq? n 2)) :yes
+                _ :no) ==> :yes ]
+       [ (match [1 2]
+                ['x 'x] :no
+                ['x 'y] :yes) ==> :yes ]
+       [ (match [1 1]
+                ['x 'x] :yes
+                ['x 'y] :no) ==> :yes])
```
jkarni: Can we have a test for the behavior of a `match` that doesn't match anything? (I.e., right exception thrown)

rad/prelude/patterns.rad
```diff
@@ -0,0 +1,192 @@
...
+      [:Just merge]
+      (maybe-foldlM (fn [acc kv]
+                      (def v (nth 1 kv))
+                      (def v0 (nth 0 v))
+                      (if (eq? v0 (nth 1 v))
+                        [:Just (insert (nth 0 kv) v0 acc)]
+                        :Nothing))
+                    merge
+                    (seq cs)))))
+(def _
```
jkarni: Hm, do we need this? If `_` remains a normal identifier, don't we get exactly the same behavior (modulo `_` actually being bound)?

rad/prelude/patterns.rad
```diff
@@ -0,0 +1,192 @@
...
+       [ (match 42 12 :not-42 42 :yes) ==> :yes ]
+       [ (match :fourty-two :not-42 :no :fourty-two :yes) ==> :yes ]
+       [ (match "42" "?" :no "42" :yes) ==> :yes ]
+       [ (match {:a :a :b :b} {:a 'a :b 'b} [a b]) ==> [:a :b] ]
+       [ (match [1 2 3] ['x _ 'y] [x y]) ==> [1 3] ]
+       [ (match {:a 1 :b [1 2 3]}
+                42 :no
+                ({:a _ :b [_ 'n _]} | (eq? n 2)) :yes
+                _ :no) ==> :yes ]
+       [ (match [1 2]
```
jkarni: What about when the thing being matched is an identifier? Shouldn't that also match a (double-quoted here) identifier?

rad/prelude/patterns.rad
```diff
@@ -0,0 +1,192 @@
...
+    (if (atom? var)
+      (fn [v]
+        (maybe->>= (match-pat pat v)
+                   (fn [bs] [:Just (insert var v bs)])))
+      (throw 'pattern-error "The first argument of `/as` should be an atom."))))
+
+(:test "As-patterns work"
+       [ (match [1 2 3]
+                (/as 'xs [_ 7 _]) :no
+                (/as 'xs [_ _ 3]) xs) ==> [1 2 3]])
```
jkarni: Can we have a test for a nested `as`? E.g:
```
       [ (match [1 2 [3 4]]
                (/as 'xs [_ 7 _]) :no
                ([_ _ (/as 'xs [_ _]]) xs) ==> [3 4]])

```

src/Radicle/Internal/Core.hs
```diff
@@ -503,6 +504,50 @@ specialForms = Map.fromList $ first Ident <$>
...
+      v : cases -> do
+        cs <- evenArgs "match" cases
+        v' <- baseEval v
+        goMatches v' cs
+      _ -> throwErrorHere $ OtherError "Pattern match must be given a value to match"
+
+    goMatches _ [] = throwErrorHere $ OtherError "Pattern match failure."
+    goMatches v ((m, body):cases) = do
+      let (m', guard_) = case m of
+            List [x, Atom (Ident "|"), g] -> (x, Just g)
```
jkarni: Aren't guards/`|` now moot? (I don't see any actual usage of `|`)

src/Radicle/Internal/Core.hs
```diff
@@ -503,6 +504,50 @@ specialForms = Map.fromList $ first Ident <$>
...
+            _                             -> (m, Nothing)
+      m'' <- baseEval m'
+      matchPat <- lookupAtom (Ident "match-pat")
```
jkarni: Lol.

jkarni:
> Are we okay with quoted variables in patterns?

I'm definitely okay with this - there's a very good reason for it, as we can now have first-class patterns.

rad/prelude/patterns.rad
jkarni: Oh yeah. Then can we maybe name it `\_`? That'd make it clear it's pattern-specific, and leave `_` for other uses.

jkarni: Hm, actually, that might be quite a bit uglier, nvm.

rad/prelude/patterns.rad
```diff
@@ -0,0 +1,192 @@
...
+
+(def /Just
+  "Pattern which matches `[:Just x]`."
+  (fn [pat]
+    (fn [v]
+      (maybe->>= v
+                 (fn [x] (match-pat pat x))))))
+
+(:test "Just patterns"
+       [ (match [:Just 42] (/Just 'x) x _ :no) ==> 42 ]
```
jkarni: I'm still not clear on why we've settled on `:Just`/`:Nothing` over `:just`/`:nothing`, but since we have already, I guess we can talk about that in some other occasion.

src/Radicle/Internal/Core.hs
jkarni: I think so. We can figure out whether we need them again later.

rad/prelude/patterns.rad
```diff
@@ -0,0 +1,192 @@
...
+      (if (list? v)
+        (if (> (length v) 0)
+          (maybe->>=
+           (match-pat x-pat (head v))
+           (fn [b1]
+             (maybe->>=
+              (match-pat xs-pat (tail v))
+              (fn [b2] (non-linear-merge b1 b2)))))
+          :Nothing)
+        :Nothing))))
```
jkarni: For when we have ident-manipulation:

```
(def test-pat x
   (match (eval x)
      (over (@nth 0) (fn [con] (cons `\` con)) x) :matches)
      _ :doesnt-match)))

(:test "\cons" [ (test-pat '(cons 1 2)) ==> :matches) ] )
```

rad/prelude/patterns.rad
jameshaydon: Actually, a related question is the syntax for pattern functions. I have chosen a prefix `\` here, but I also considered using capital letters: `Just`, `Cons`, etc. This would make it consistent with the use of capital letters when mimicking "constructors" in Haskell.

rad/prelude/patterns.rad
jameshaydon: That would involve adding a case for `(quote x)` in the dispatch function, which I'm not sure is a good idea, because we might want to use lists for matching lists or something. People can write a pattern for matching atoms if they want.

rad/prelude/patterns.rad
jkarni: I slightly prefer titlecase, but up to you.

src/Radicle/Internal/Identifier.hs
```diff
@@ -22,7 +22,7 @@ isValidIdentRest x = isAlphaNum x || x `elem` extendedChar
 
 extendedChar :: Prelude.String
 extendedChar = ['!', '$', '%', '&', '*', '+', '-', '.', '/', ':', '<' , '=', '>'
-  , '?', '@', '^', '_', '~']
+  , '?', '@', '^', '_', '~', '|']
```
jkarni:
```suggestion
  , '?', '@', '^', '_', '~']
```

cloudhead:
Amazing!

A couple of questions:
* Does this work with erlang style pattern matches like `{A, A} -> ...` which effectively means `{A, B} | A == B -> ...` Or would that be a binding error?
* Does the Maybe type make sense in an untyped language? Wouldn't you just pass `nil` instead?
* Is there a reason why we're capitalizing `:Just/:Nothing`? Are they supposed to be distinguished from other keywords?

jkarni:
> Does this work with erlang style pattern matches like {A, A} -> ... which effectively means {A, B} | A == B -> ... Or would that be a binding error?

Yup, it handles these (non-linear) patterns. I'll add a couple of tests.

> Does the Maybe type make sense in an untyped language? Wouldn't you just pass nil instead?

The general idea of maybe I think carries over decently well to untyped languages. E.g., if lookup fails with nil, you can never distinguish a dictionary having a key that maps to nil from the lookup having failed, whereas if you use this pattern you get `[:Just nil]` in the one case and `:Nothing` in the other. I feel like in Erlang people do this quite a bit too?

> Is there a reason why we're capitalizing :Just/:Nothing? Are they supposed to be distinguished from other keywords?

Agreed. We've been using this already, so I'm in favor of doing the work of switching to lowercase in another PR, but I also don't see much reason to capitalize (and we can then keep capitalization in our pockets for some future distinction).

PR #237, opened 2018-11-22
Handles, processes, IO

jkarni:
This PR introduces handle- and process-related functionality to radicle that I wanted as part of the process to make nicer CLI interfaces. There's still some work to be done to make this more ergonomic, such as reading all contents from a handle and concurrency (which is also necessary for proper tests, since otherwise reading stdout/stderr is difficult). But perhaps since this touches a lot of core, it makes sense to separate out PRs to avoid merge conflicts (happy to continue if people think otherwise).

Other things this PR does:

- Add `put-str!` and define `print!` in terms of it and `show`
- Incorporate Thomas' PR avoiding `To/FromRad` with bindings [de]serialization
- Add some string processing functions.
- Start on utils (one of which isn't ready yet)


rad/prelude/io.rad
```diff
@@ -22,3 +22,20 @@
...
+(def shell!
+  "(shell! command stdin) executes 'command' using the shell with 'stdin' as
+   input. Returns the exit code.
+   Example: (shell! \"ls -Glah\" \"\")
+   "
+  (fn [command stdin]
+    (system! [:shell command] stdin)))
+
+(def process!
+  "(process! command args stdin) executes 'command' using 'execvp'. with 'stdin' as
```
geigerzaehler: It would be nice to explain what `execvp` is to the non-C programmers ;)

src/Radicle/Internal/PrimFns.hs
```diff
@@ -257,6 +251,16 @@ purePrimFns = fromList $ allDocs $
             pure . Dict . Map.fromList $ zip (fst <$> kvs) vs
           (_, v) -> throwErrorHere $ TypeError "map-values" 1 TDict v
       )
+    , ( "map-keys"
+      , "Given a function `f` and a dict `d`, returns a dict with the same values as `d`\
```
jameshaydon: Maybe say what happens if `f` is not injective.

src/Radicle/Internal/PrimFns.hs
```diff
@@ -257,6 +251,16 @@ purePrimFns = fromList $ allDocs $
...
+        \ but `f` applied to all the keys."
+      , twoArg "map-keys" $ \case
+          (f, Dict m) -> do
+            let kvs = Map.toList m
+            vs <- traverse (\v -> callFn f [v]) (fst <$> kvs)
+            pure . Dict . Map.fromList $ zip vs (snd <$> kvs)
```
jameshaydon: Use `dict` because `f` might produce non-hashable keys.

src/Radicle/Internal/Effects.hs
```diff
@@ -208,6 +209,50 @@ replPrimFns = fromList $ allDocs $
...
+          let ifJustCreate x = case x of
+                Nothing -> pure Nothing
+                Just v  -> Just <$> newHandle v
+          case res of
+              (a, b, c, ph) -> do
+                  a' <- ifJustCreate a
+                  b' <- ifJustCreate b
+                  c' <- ifJustCreate c
+                  ph' <- newProcessHandle ph
+                  pure $ toRad (a', b', c', ph')
```
jameshaydon: It seems to return more than just the exit code?

src/Radicle/Internal/Effects.hs
```diff
@@ -208,6 +209,50 @@ replPrimFns = fromList $ allDocs $
...
+      )
+    , ( "write-handle!"
+      , ""
```
jameshaydon: docs

src/Radicle/Internal/Effects.hs
```diff
@@ -208,6 +209,50 @@ replPrimFns = fromList $ allDocs $
...
+      , twoArg "write-handle!" $ \case
+          (Handle h, String msg) -> do
+              h' <- lookupHandle h
+              hPutStrS h' msg
+              pure $ Keyword $ Ident "ok"
+          (Handle _, v) -> throwErrorHere $ TypeError "write-handle!" 1 TString v
+          (v, _) -> throwErrorHere $ TypeError "write-handle!" 1 THandle v
```
jameshaydon:
```suggestion
          (v, _) -> throwErrorHere $ TypeError "write-handle!" 0 THandle v
```

src/Radicle/Internal/PrimFns.hs
jameshaydon: http://hackage.haskell.org/package/containers-0.6.0.1/docs/Data-Map-Strict.html#v:traverseWithKey

rad/prelude.rad
```diff
@@ -1,11 +1,13 @@
 (load! "rad/prelude/test.rad")
 (load! "rad/prelude/basic.rad")
 (load! "rad/prelude/patterns.rad")
+(load! "rad/prelude/strings.rad")
 (load! "rad/prelude/io.rad")
 (load! "rad/prelude/bool.rad")
 (load! "rad/prelude/exception.rad")
 (load! "rad/prelude/recursion.rad")
 (load! "rad/prelude/list.rad")
+(load! "rad/prelude/strings.rad")
```
jameshaydon: Added twice.

rad/prelude/strings.rad
```diff
@@ -0,0 +1,31 @@
...
+(def lines
+  "Concatenate a list of strings, with newlines in between."
+  (fn [x]
+    (intercalate "\n" x)))
+
+(:test "lines"
+  [ (lines ["hi" "there"]) ==> "hi\nthere" ]
+)
+
+(def words
```
jameshaydon: `unwords`?

rad/prelude/io.rad
```diff
@@ -4,17 +4,49 @@
...
+  "(process! command args to-write) executes `command` using 'execvp'. with 'to-write' as input. Returns the exit code.  See 'man exec' for more information on 'execvp'.  Example: (process! \"ls\" [\"-Glah\"] \"\") "
+  (fn [command args to-write]
+    (def cp
+      { :cmdspec [:raw command args]
+        :stdin :create-pipe
+        :stdout :inherit
+        :stderr :inherit
+      })
+    (match (system! cp)
+      [[:just 'stdin] :nothing :nothing 'ph] (do (write-handle! stdin to-write)
```
jameshaydon: duplicate code

src/Radicle/Internal/Core.hs
```diff
@@ -375,11 +441,77 @@ instance GhcExts.IsList (PrimFns m) where
...
+--
+-- prop> gets bindingsToRadcile >>= setBindings = pure ()
+bindingsToRadicle :: Bindings a -> Value
+bindingsToRadicle x =
+    Dict $ Map.fromList
+        [ (Keyword $ Ident "env", env)
+        , (Keyword $ Ident "refs", refs)
+        ]
+  where
+    env = Dict . Map.mapKeys Atom . Map.map toRad . fromEnv $ bindingsEnv x
```
jameshaydon: Map.mapWithKey

jkarni: Hm, not sure I get it. We're mapping the keys and the values, but mapWithKey takes the key and the value as arg but only changes the value

jameshaydon: Indeed ignore me

rad/prelude/io.rad
jameshaydon: With modules this problem will go away, so wait for that.

geigerzaehler:
Can we rebase and squash these kinds of PRs before merging them? Having a bunch of `wip`s in your commit history is not helpful.

jkarni:
> Can we rebase and squash these kinds of PRs before merging them? Having a bunch of wips in your commit history is not helpful.

Yeah, sorry, the default is usually that, but I guess I switched it at some point and didn't notice. @geigerzaehler  and @jameshaydon can I force push something to squash these in master?

jameshaydon:
> can I force push something to squash these in master

np

PR #243, opened 2018-11-23
Better error messages

jameshaydon:


src/Radicle/Internal/Pretty.hs
```diff
@@ -23,6 +26,9 @@ instance Pretty Reference where
 instance Pretty Ident where
     pretty (Ident i) = pretty i
 
+instance Pretty Type where
+  pretty = fromString . drop 1 . show
```
jkarni: Should this use `typeToValue`?

exe/RadicleExe.hs
```diff
@@ -16,6 +16,7 @@ import           System.Directory (doesFileExist)
 
 import qualified Radicle.Internal.Number as Num
 import qualified Radicle.Internal.PrimFns as PrimFns
+import           Radicle.Internal.Type (Type(..))
```
jkarni: I'd re-export `Type` from `Radicle` instead.

src/Radicle/Internal/Pretty.hs
jameshaydon: Yes this and `typeToValue` will both just use https://github.com/oscoin/radicle/pull/242/files#diff-dd5f436b10a11b14706cef020de15c04R44 when that is merged and I rebase this PR.

Issue #238, opened 2018-11-22
Replace the term 'chain' with 'machine'


geigerzaehler:
We use the term *chain* throughout the docs and code. The term incorporates a couple of concepts
* a list of radicle expressions
* an associated state
* something that can be updated
* something that can be hosted
* something that can be identifie4d

In my opinion, *chain* does not describe this adequately and is moreover confusing to people from the outside. I would propose using the term *machine* (short for *Radicle state machine*) everywhere we use chain.

cc @clevinson

cloudhead:
:+1:

clevinson:
I'm in favor of replacing "chain" with something that better reflects its uses. Chain is used too often as short-hand for blockchain within the larger p2p / crypto community. However, I see similar issues with "machine", as this usually refers to physical hardware.

Are there other ideas for terms we could use to abbreviate a Radicle RSM? Do people like the name "feed" ?

hxrts:
I've run into this while explaining the product to a few people. I think highlighting the intrinsic connection between the two can actually be quite useful (at least early on) i.e. "machine" is the operational form and "chain" is the product that represents an ordered record of all operations.

cloudhead:
@clevinson "feed" only captures the *input* part of the rsm, not the whole thing.

cloudhead:
How about this:

* **rsm**: the main underlying concept. Includes expressions, states and a head.
* **expression**: an input to the rsm - the primitive in our network.
* **radicle network**: the set of all rsms.
* **state**: materialized view on an rsm.
* **app**: a user-level materialization of one or more **rsm**s.

geigerzaehler:
I agree Radicle state machine should be the full name that disambiguates it from a user’s host machine. But I’d shorten it to “machine” instead of “rsm”. It makes the documentation much easier to read, for example

> Now we can start a radicle REPL that will be able to communicate with the remote machine.

versus

> Now we can start a radicle REPL that will be able to communicate with the remote RSM.

The rest of the terms you proposed, @cloudhead, makes total sense.

I would also like to keep this concrete instead of abstract, though. What I want to address is confusion for outsiders in the documentation. So please take a look at PR #249 and see if the terms you proposed make sense there. I updated it to use the term “app” in the tutorial to build an issues app (instead of an issues chain).

cloudhead:
[...]

PR #249, opened 2018-11-26
Replace 'chain' with 'machine' in docs

geigerzaehler:
To make the discussion in #238 more concrete we replace all occurences of the term “chain” with “machine” in the documentation.


geigerzaehler:
Most of this was done in #415

PR #250, opened 2018-11-26
Add `radicle-counter` example app

geigerzaehler:
We add a executable that can be used to control a simple counter application on radicle.

This app can be used to test chain storage and replication.

### DIFF
--- a/examples/radicle-counter
+++ b/examples/radicle-counter
@@ -1,0 +1,98 @@
+#!/bin/sh
+
+usage() {
+  cat <<USAGEDOC
+USAGE
+  $0 <chain> <cmd>
+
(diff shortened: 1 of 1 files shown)


examples/radicle-counter
```diff
@@ -0,0 +1,98 @@
...
+  exit 1
+fi
+
+case "$cmd" in
+  init )
+    radicle - <<-SCRIPT
+      (load! "rad/prelude.rad")
+      (send! "${chain}" '[
+        (def counter (ref 0))
+        (def increment!
```
jkarni: I think there shouldn't be an `!` here; we've been using `!` for *IO* effects, not ref/env ones.

PR #254, opened 2018-11-28
Add 'weeder' to CI

geigerzaehler:
Add [weeder][1] to our CI pipeline. Weeder checks for unused dependencies and exports. This forces us to remove some unused dependencies.

[1]: https://github.com/ndmitchell/weeder


geigerzaehler:
> Nice! But it doesn't look like CI is running this?

oops

PR #257, opened 2018-11-28
Adds modules

### LINKED ISSUE #232: Make ref-doc use text rather than source file for layout etc
Currently any documentation change involves a recompilation, which is quite annoying. Additionally, the source could be more readable.

jameshaydon:
This PR:
- Adds first-class modules.
- Makes the reference doc use module-documentation, as well as the order specified by the module export vectors. Other aspects of the reference doc are now specified in a YAML file. No more tweaking `ReferenceDoc.hs`! :champagne: 
- Adds module declarations to all the radicle source files, and the necessary import statemenets.

- [x] Add tests
- [x] ReferenceDoc should use docstrings in modules, and the order given by the export list.
- [x] Add a guide on modules.

Closes #232


rad/chain.rad
```diff
@@ -1,3 +1,12 @@
+'prelude/chain
```
jkarni: So a top-level ident defines the module name?

rad/monadic/issues.rad
```diff
@@ -4,7 +4,24 @@
 ;; Create a remote issue chain. Returns a ref for the chain that you
 ;; can assign to a variable.
 
-(load! "rad/prelude.rad")
+(load! "rad/prelude/test.rad")
```
jkarni: Is it no longer possible to have prelude re-export a bunch of things?

rad/prelude/bool.rad
```diff
@@ -1,3 +1,9 @@
+'prelude/bool
+
+"Functions for dealing with truthiness and #f."
```
jkarni: Not sure if at the RFC stage I should be talking only about broader points than syntax, but 

```
(module 'prelude/bool '[not and or all]
  "Functions for ...") 
```
Looks prettier and less magical to me.

rad/monadic/issues.rad
jameshaydon: No it is, I was just debugging this file and wanted to see exactly what was getting imported. (Because if you run it from normal repl, it would load prelude twice, etc.) I can change it back.

rad/prelude/bool.rad
jameshaydon: Yeah I agree, but I used `module` as the special-form to create modules inline. Also I want to keep files executable even if they aren't being treated as modules. So e.g.:

```clojure
`[:module prelude/bool [not and or all]
  "Functions for ..."]
```
?

jkarni: I guess we could just define a no-op module function, but I actually like the vector better.

jkarni: Hm, in the interest of making it easier to have backwards compatibility, maybe a dict? E.g.:
```
{:module 'prelude/bool
 :doc "Functions for ..."
 :exports '[not and or all]
}

File: src/Radicle/Internal/Core.hs [hunk 4]
```diff
@@ -601,6 +589,26 @@ specialForms = Map.fromList $ first Ident <$>
...
+createModule :: Monad m => Text -> [Value] -> Lang m Value
+createModule ctx = \case
+    (name : doc : exports : forms) -> do
+      name' <- baseEval name
+      doc' <- baseEval doc
+      exports' <- baseEval exports
+      case (name', doc', exports') of
+        (Atom n, String d, Vec is) -> do
+          is' <- traverse isAtom is ?? toLangError (OtherError $ ctx <> ": Module export vector should only contain atoms")
+          e <- withEnv identity $ traverse_ eval forms *> gets bindingsEnv
```
jkarni: Is this right? Shouldn't it be `withEnv [im]pureEnv`? I'd think that modules are evaluated in a blank environment, so that if they "work" in one place they always work.

jkarni: (Sorry, I meant `withEnv (const pureEnv)`)

jameshaydon: - The current way a module uses code from another module is by `(import another)` where usually `another` is an identifier referring to a previously `def`ined module. So if the module was ran in a pure env, it wouldn't find this. If a module is to start from a totally pure env, I don't see how we can avoid running modules it depends on multiple times (if e.g. `a` and `b` both depend on `c`).

- Modules mostly assume the env has already been changed, e.g. `(:test ...)` assumes the `eval` has been changed.

- I quite like the dynamic nature. E.g. I can do:

```clojure
(def counter
  (fn [starts-at]
    (def i (ref starts-at))
    (def next ...)
    (def current ...)
    (module '[next current])))

...

(import (counter 42) issue-id)
(def my-issue-id (issue-id/next))
```

geigerzaehler:
It would be very helpful to have a section in the documentation detailing how the module system is supposed to work.

docs/source/guide/Modules.lrad
```diff
@@ -0,0 +1,70 @@
+# Modules
+
+Modules are dicts with some metadata about a module, and an `:env` key
+containing a radicle environment.
```
geigerzaehler: Can we move this description of the implementation somewhere to the bottom. We could lead with a more generic description of what modules are.

docs/source/guide/Modules.lrad
```diff
@@ -0,0 +1,70 @@
...
+       :exports '[foo bar]}
+
+      (def foo-helper
+        (fn [x] (+ x 1)))
+
+      (def foo
+        (fn [x] (foo-helper (foo-helper x))))
+
+      (def bar
+        (fn [x] (foo (+ x 2)))))
```
geigerzaehler: Could you explain how atoms that are not defined in the module are resolved?

geigerzaehler: I’d actually be very interested in having parameterized modules.
~~~clojure
(module
  {:name 'foo
   :parameters {:base 0}
  }
  (def add (fn [x] (+ base x))
)

(import foo-default foo)
(foo-default/add 1) ;; => 1

(import foo-custom foo {:base 3})
(foo-custom/add 1) ;; => 4

jameshaydon: That's mostly achievable already just using lambdas:

```clojure
(def mk-foo
  (fn [base]
    (module
     {:module 'foo
      :doc "foo"
      :exports ['add]}
     (def add (fn [x] (+ base x))))))
```

and then:

```
rad> (import (mk-foo 0))
:ok
rad> (foo/add 0)
0
rad> (import (mk-foo 3))
:ok
rad> (foo/add 0)
3
```

docs/source/guide/Modules.lrad
```diff
@@ -0,0 +1,107 @@
+# Modules
+
+To create a module, use the `module` special form. All modules must start with a
+*module declaration* which provides the module with a name, documentation string
+and an exports vector. Module declarations are evaluated, so you must quote any
```
jkarni: `any ...?` Maybe:
```suggestion
and an exports vector. Module declarations are evaluated, so you must quote any exports.
```

docs/source/guide/Modules.lrad
```diff
@@ -0,0 +1,107 @@
...
+
+You can narrow down exactly which definitions you would like to import by
+including a list of atoms:
+
+    (import my-module ['foo])
+
+Which would only import `my-module/foo`. And you can also use a custom
+qualifier:
+
+    (import my-module 'baz)
```
jkarni: Would it be a lot of work to have clojure-style qualified imports (`(import my-module :as 'baz)`)? I imagine people expecting `(import my-module 'baz)` to import `baz` alone.

rad/monadic/issues.rad
```diff
@@ -4,20 +4,37 @@
...
+(file-module! "rad/prelude/io.rad")
+(import prelude/io :unqualified)
+
+(file-module! "rad/prelude/ref.rad")
+
+(file-module! "rad/prelude/dict.rad")
+(import prelude/dict :unqualified)
+
+(file-module! "rad/prelude/lens.rad")
+(import prelude/lens :unqualified)
```
jkarni: Any reason to not just import prelude completely?

rad/monadic/issues.rad
```diff
@@ -4,20 +4,37 @@
...
+(import prelude/chain 'chain)
 
 (def radicle-issues-chain-id
   "The name of this chain."
   "monadic/radicle/issues")
 
-(def issues-chain!
-  (ref (load-chain! "http://radicle.xyz/chains/radicle-issues")))
+(def issues-chain
+  (ref (chain/load-chain! "localhost:8000/chains/radicle-issues")))
```
jkarni: Hm. Maybe parametrize the module by url? @geigerzaehler were just talking about this in relation to the tests.

rad/monadic/issues.rad
```diff
@@ -56,13 +73,13 @@
 (def create-issue!
   "Create a new remote issue with the keys in `my-keys.rad`."
   (fn [i]
-    (send! (lookup :url (read-ref issues-chain!))
+    (send! (lookup :url (read-ref issues-chain))
```
jkarni: Seems like a lot of `!` are getting lost.

src/Radicle/Internal/Core.hs
```diff
@@ -629,6 +624,10 @@ specialForms = Map.fromList $ first Ident <$>
               _ -> throwErrorHere $ SpecialForm "fn" "First argument must be a vector of argument atoms"
           _ -> throwErrorHere $ SpecialForm "fn" "Need an argument vector and a body"
       )
+  , ("scope", \case
```
jkarni: This doesn't seem to be used anywhere? And isn't it essentially a self-calling lambda?

src/Radicle/Internal/PrimFns.hs
```diff
@@ -531,6 +545,34 @@ purePrimFns = fromList $ allDocs $
           (x@(Atom _), v) -> pure $ toRad (Just (Dict (Map.singleton x v)))
           (pat, v) -> callFn pat [v]
       )
+    -- , ( "module-from-env"
```
jkarni: Delete?

rad/monadic/issues.rad
jameshaydon: These functions are not effectful so I removed the `!`.

rad/monadic/issues.rad
jameshaydon: Not sure what's best here so I'll leave it to a later PR.

jkarni: Sounds good.

rad/monadic/issues.rad
jameshaydon: `issue-chain` *is* a ref, should refs have `!`?

The other functions don't have effects because all they do is read a ref (which I thought we agreed was not an effect), the actual effectful stuff is all concentrated in `fetch!`.

docs/source/guide/Modules.lrad
```diff
@@ -0,0 +1,107 @@
...
+        (fn [x] (foo-helper (foo-helper x))))
+
+      (def bar
+        (fn [x] (foo (+ x 2)))))
+
+To import a module, use `import`:
+
+    (import my-module)
+
+The definitions inside a module are evaluated in the current scope, so you can
```
geigerzaehler: Sorry to be picky but I think it’s important to be precise here to avoid confusion: Can we change this to

> The definitions inside a module are evaluated in the scope of the module definition site.

It would also be helpful to have another example like this

~~~clojure
(def foo :foo)
(module
  { ... }
  (def bar foo))

rad/monadic/issues.rad
jkarni: OK, so @jameshaydon convinced me on chat that not having a `!` is reasonable.

PR #258, opened 2018-11-28
Build `radicle-server` image in CI

geigerzaehler:
We build and push the `eu.gcr.io/opensourcecoin/radicle-server` image that runs the radicle server.

To achieve this we need a bit of trickery when using `stack image container`.

We also optimize the CI build a bit:
- Parallelize the jobs after “Build & Test” with `waitFor` key
- Silence `apt-get` with `-qq` flag
- Remove superfluous `postgresql` dependency

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -18,3 +18,4 @@ docs/build/
 docs/build/
 TAGS
 rad/my-keys.rad
+/images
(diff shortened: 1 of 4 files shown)


geigerzaehler:
So, the image does not really work. I’ll close this PR on follow-up with a new one.

PR #259, opened 2018-11-29
Build a working image of radicle-server

geigerzaehler:
We add the capability to build a working `radicle-server` image locally and on CI. We push the image to `eu.gcr.io/opensourcecoin/radicle-server` on CI.

The previous setup using Stack’s [`image` configuration][1] did not work because the image requires `libpq`. Now we add the `images/radicle-server` folder that holds the `Dockerfile` and scripts
to build the image locally and on CI.

Note that for local builds with the `build.sh` script we tell Stack to build the binaries inside a Docker conatiner (using `stack build --docker`). This ensures that the produced binary is compatible with the
distribution in the base image.

[1]: https://docs.haskellstack.org/en/latest/GUIDE/#docker


images/radicle-server/Dockerfile
```diff
@@ -0,0 +1,16 @@
+FROM debian:stretch
+# The distribution must be the same one the binary is build with to
+# ensure binary compatibility. We use the `haskell` images on CI and
+# `fpco/stack-build` locally.
```
jkarni: What is this referring to? (I mean using `fpco/stack-build` locally.)

images/radicle-server/Dockerfile
```diff
@@ -0,0 +1,16 @@
+FROM debian:stretch
```
jkarni: Why not also use `haskell:8.4.3`?

exe/Server.hs
```diff
@@ -195,15 +199,22 @@ opts = mkOpts
...
+    mkOpts h pgp u pw db = Opts (ConnectInfo h pgp u pw db)
 
 
 -- * Main
 
 main :: IO ()
 main = do
     opts' <- execParser allOpts
+    when (optsShowVersion opts') $ do
+        putStrLn $ "radicle-server version " <> showVersion version
```
jkarni: Is this useful? We haven't changed cabal-file versions at all so far, and if we're generating this for each new commit on master, we definitely are going to have lots of *different* binaries with the same version even if we do start bumping sometimes.

geigerzaehler: I added this to have something to run to check that `radicle-server` is correctly linked when executed in the container (see https://github.com/oscoin/radicle/pull/259/files#diff-889dc80f39676827dc606c84c4baf1e0R13). However I realize that we can just use `radicle-server --help >/dev/null`. I’ll make the change.

images/radicle-server/Dockerfile
geigerzaehler: This refers to building the binary with `stack build --docker`. I added that to the documentation.

PR #271, opened 2018-12-04
Reorganize self testing in Radicle

geigerzaehler:
We change how the `:test` macro is implemented. Instead of running a test immediately when it is encountered we register it in the environment and provide the `run-all-tests` function. The Haskell tests then call the functions.

### DIFF
--- a/rad/prelude/test-eval.rad
+++ b/rad/prelude/test-eval.rad
@@ -1,0 +1,58 @@
+{:module 'prelude/test-eval
+ :doc
+   "Provides eval that adds a `:test` macro. If an the eval encounters an
+   expression of the form `(:test \"desc\" tests...)` we add it to the vector
+   stored in the `tests` ref. The tests are then run by `rad/prelude/test.rad`."
+ :exports '[eval tests]}
+
+(def tests
+  "Reference that collects tests defined with the `:test` macro. This
+  value is updated by a custom `eval`."
+  (ref []))
+
+;; Annoyingly verbose since we have neither the prelude nor pattern-matching.
+(def eval
+  "Evaluation function that adds :test macro to register tests."
+  (fn [expr env]
+    (if (list? expr)
+        (if (eq? (head expr) :test)
+            (do
+                (def bindings (lookup :env env))
+                (def test-def {:tests (drop 2 expr) :name (nth 1 expr) :env env})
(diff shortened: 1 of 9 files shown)


rad/prelude/test.rad
```diff
@@ -1,73 +1,48 @@
-;; testing functions
+;; This file provides the `run-all-tests` function that runs the tests
+;; registered with the `:test` macro.
 
-(def is-test-env
-  "True iff file is being run as part of the Haskell suite"
-  (catch 'any (eq? test-env__ #t) (fn [x] #f)))
+(def run-test
+  "Run `test` and print the result. `test` is is an list of quoted
```
jkarni:
```suggestion
  "Run `test` and print the result. `test` is a list of quoted
```

rad/prelude/test.rad
```diff
@@ -1,73 +1,48 @@
...
+  )
+  ```
+  "
+  (fn [test]
+    (def test-name (nth 0 test))
+    (def error
+      (fn [msg]
+        (put-str! (string-append "Test '" test-name "' failed: " msg))))
+    ;; Check a single test in a given env
+    (def check-single
```
jkarni: Maybe this is for a different PR, but it'd be nice if `run-test` caught exceptions - currently we get really weird failures if any test throws, and it's hard to figure out exactly what threw them. (Also, no *later* tests are run.)

geigerzaehler: https://github.com/oscoin/radicle/issues/274

geigerzaehler:
@jkarni: I addressed the issues we discussed offline

rad/prelude/test-eval.rad
```diff
@@ -0,0 +1,58 @@
...
+ :exports '[eval tests]}
+
+(def tests
+  "Reference that collects tests defined with the `:test` macro. This
+  value is updated by a custom `eval`."
+  (ref []))
+
+;; Annoyingly verbose since we have neither the prelude nor pattern-matching.
+(def eval
+  "Evaluation function that adds :test macro to register tests."
```
jameshaydon: The annoying thing here is in the repl (and the reference doc), the doc for `eval` now mainly talks about tests.

geigerzaehler: I’ll fix this in an upcoming PR.

Issue #196, opened 2018-11-05
Further unify vec and list operations


jkarni:
- `cons` and/or `add-left` could operate on both types
- `<>` could operate on both types

jkarni:
The `head` and `tail` part of this were done in #198.

PR #277, opened 2018-12-04
Make `cons` and `filter` work for vectors

geigerzaehler:


jkarni:
> I'm just wondering what we should do about sequence polymorphism in general. Now we have cons and add-left which do the same thing except add-left is just for vectors. It seems a bit arbitrary.

Agreed - like Clojure's [`seq`](https://clojure.org/reference/sequences). I'd say `String` should also have `cons` and `filter`.

geigerzaehler:
How do we proceed here? Should I make some change or should I close this PR?

jameshaydon:
We should have `head`, `tail` and `cons` functions that work on all sequences (as this PR implements). Not sure about the names though, we might want names more like Clojure (`first`, `rest`, `cons`) to imply that it works on all sequences, not just lists.

Another thing is that whatever happens, the patterns should match the functions, that is, `/cons` in `prelude/patters.rad` should be made to work on lists and vectors if that's what `cons` does.

geigerzaehler:
How about this: I’ll make `/cons` work with Vectors, too and add `first` and `rest` to work with both vectors and lists.

jameshaydon:
Sounds good, and then close #196.

geigerzaehler:
@jameshaydon: Can you have another look?

jameshaydon:
Shall we also get rid of `add-left`?

geigerzaehler:
> Shall we also get rid of `add-left`?

Done

PR #281, opened 2018-12-05
Don’t check reference doc rst into source

geigerzaehler:
We remove `docs/source/reference.rst` from version control so that we don’t need to remember to re-generate it before we submit a PR.


jameshaydon:
But don't we need it in the repo for http://docs.radicle.xyz/en/latest/?

geigerzaehler:
> But don't we need it in the repo for http://docs.radicle.xyz/en/latest/?

Indeed. Read The Docs uses webhooks and fetches the source code. I’ll think about how to improve this.

geigerzaehler:
It does not seem possible to generate this file during build time.

PR #286, opened 2018-12-05
Move counter app code into rad file

geigerzaehler:
We move the code for the counter example app from the Bash script to its own Radicle source file. We also provie tests for the counter app.

### DIFF
--- a/rad/examples/counter.rad
+++ b/rad/examples/counter.rad
@@ -1,0 +1,64 @@
+;; A “counter” chain that exposes an `increment` mutator and a `get-value`
+;; accessor.
+;;
+;; This file provides functions to interact with a counter chain.
+;;
+;;     (counter/init-chain! "my-chain-id") ;; => :ok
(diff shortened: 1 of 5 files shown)


rad/examples/counter.rad
```diff
@@ -0,0 +1,64 @@
...
+;;     (counter/get-value "my-chain-id") ;; => 0
+;;     (counter/increment! "my-chain-id") ;; => 1
+;;     (counter/get-value "my-chain-id") ;; => 1
+;;
+
+(load! "rad/prelude.rad")
+
+(import prelude/test '[assert-equal] :unqualified)
+
+(def counter/prelude
```
jkarni: I think now that we have modules and qualified imports this doesn't need the leading `counter/`?

geigerzaehler: Removing the `counter/` prefix would indeed make sense if this file was a module. But making this file a module would require more boilerplate code or another wrapper in the bash script. This is why I opted for this file to be just a script.

PR #287, opened 2018-12-06
Make QuasiQuoters work properly

geigerzaehler:
The quasi quoters `ident` and `kword` now throw, as advertised, at compile time. They also throw if there is a space in the identifier. We also move them into a separate module.


geigerzaehler:
@jameshaydon: Could you have another look? I had to make some changes to prevent a regression

jameshaydon:
@geigerzaehler :+1:

PR #288, opened 2018-12-06
Better code and errors for reference docs

geigerzaehler:
The code the generates the reference docs now
* provides better error messages
* and does not rely on hardcoded values in `checkAllInReference` anymore.

The `eval` documentation is now removed, which I think is ok given the fact that it only mentioned the `:test` macro.

### DIFF
--- a/docs/source/reference.rst
+++ b/docs/source/reference.rst
@@ -87,11 +87,6 @@ The default evaluation function. Expects an expression and a radicle
 state. Return a list of length 2 consisting of the result of the
 evaluation and the new state.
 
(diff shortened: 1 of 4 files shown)


jameshaydon:
So what happens if I go `(doc 'eval)` in the repl? I think we might want to special-case this so that a general message about eval-redefinition is printed (but we can leave this to another PR).

geigerzaehler:
> I think we might want to special-case this so that a general message about eval-redefinition is printed (but we can leave this to another PR).

I agree, we should handcraft the docs for this.

PR #290, opened 2018-12-06
More IO

jkarni:
Adds `shell-with-stdout!`, `process-with-stdout!`, `close-handle!`, and `read-contents-handle!`.
`fzf-select-by-key!` works again.

### DIFF
--- a/rad/prelude/io.rad
+++ b/rad/prelude/io.rad
@@ -75,8 +75,14 @@ Example: `(shell! \"ls -Glah\" \"\")`. "
             out)
       _ (throw 'io-error "expected only stdin and stdout to be open"))))
 
+(:test "shell-with-stdout!"
+  [ (shell-with-stdout! "echo hi" "") ==> ["hi"] ]
+  [ (shell-with-stdout! "cat" "there") ==> ["there"] ]
+)
+
+
 (def process!
-  "Executes `command` using 'execvp'. with 'to-write' as input. Stdout and stderr are inherit.
+  "Executes `command` using 'execvp'. with 'to-write' as input. Stdout and stderr are inherited.
 See 'man exec' for more information on 'execvp'.
 Example: `(process! \"ls\" [\"-Glah\"] \"\")`. "
   (fn [command args to-write]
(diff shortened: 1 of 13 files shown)


rad/prelude/io.rad
```diff
@@ -22,6 +26,15 @@
   (fn [x]
     (put-str! (show x))))
 
+(def-rec read-contents-handle!
+  "Read all the contents of a handle. A *vector* of strings is returned, one
```
geigerzaehler: This would read better in the documentation.
```suggestion
  "Read all the contents of `hdl`. A *vector* of strings is returned, one
```

src/Radicle/Internal/Effects/Capabilities.hs
```diff
@@ -57,23 +65,31 @@ class (Monad m) => System m where
...
 instance System IO where
     systemS = createProcess
     waitForProcessS = waitForProcess
-    hPutStrS = hPutStr
-    hGetLineS x = toS <$> hGetLine x
+    hPutStrS h t = hSetBuffering h LineBuffering >> hPutStr h t >> hFlush h
+    hGetLineS x =
+        catchJust (\e -> if isEOFError e then Just () else Nothing)
+                  (Just . toS <$> (hSetBuffering x LineBuffering >> hGetLine x))
+                  (\() -> pure Nothing)
```
geigerzaehler: You could use [`hIsEOF`](http://hackage.haskell.org/package/base-4.12.0.0/docs/System-IO.html#v:hIsEOF) instead of `catchJust`

utils/list-issues
```diff
@@ -21,9 +23,13 @@ cat << SRC > $FILE
...
      "\nState: " (if (eq? (lookup :state i) :open) "open" "closed")
      "\nLabels: " (show (lookup :labels i))
-     "\nComments:\n\n\n" (lines (map pretty-comment (lookup :comments i)))
+     "\nComments:\n\n\n" (unlines (map pretty-comment (lookup :comments i)))
 )))
 (put-str! (pretty-issue selected))
 SRC
-radicle -- $FILE | less
 
+if [ -z "$NIX_PATH" ]; then
```
geigerzaehler: This can not reliably determine if Nix is used. For example, I have `NIX_PATH` set because I have it installed, but I’m not actively using it. Isn’t there something like `USING_NIX`?

jkarni: True. It's also not clear that we should *always* be passing these extra flags - it's possible someone is using NixOS but not our particular setup.

Shall I just set a flag in `nix/shell.nix`, and then check for it here?

geigerzaehler: > Shall I just set a flag in `nix/shell.nix`, and then check for it here?

Sounds like a good idea.

src/Radicle/Internal/Effects/Capabilities.hs
jkarni: Done.

utils/list-issues
jkarni: Done.

PR #293, opened 2018-12-06
Fix basics tutorial check

MeBrei:
Add the basics tutorial code check to the build, fix the check itself and change the code snippets so that they pass.

### DIFF
--- a/exe/Doc.hs
+++ b/exe/Doc.hs
@@ -7,7 +7,6 @@ import           Protolude
 import qualified Data.Text as T
 import           Options.Applicative
 import           Radicle
-import           System.Console.Haskeline (defaultSettings, runInputT)
 import           Text.Pandoc
 
 main :: IO ()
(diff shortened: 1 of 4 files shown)


exe/Doc.hs
```diff
@@ -24,8 +24,10 @@ main = do
 run :: FilePath -> IO ()
 run f = do
     txt <- readFile f
-    pand <- runIOorExplode $ readMarkdown def txt
-    res <- runInputT defaultSettings $ runLang replBindings $ interpretMany (toS f) $ getCode pand
+    let extensions = def { readerExtensions = githubMarkdownExtensions }
+    pand <- runIOorExplode $ readMarkdown extensions txt
+    let code = T.append "(load! \"rad/prelude.rad\")" (getCode pand)
+    res <- runInputT defaultSettings $ runLang replBindings $ interpretMany (toS f) $ code
```
geigerzaehler: There’s no need to run it with `InputT`, just `runLang` should suffice

```suggestion
    res <- runLang replBindings $ interpretMany (toS f) $ code
```

PR #296, opened 2018-12-06
Sys args, open-file, and unify util scripts

jkarni:
Adds:

- `(get-args!)`
- `(open-file!)`
- `(write-file!)`
- `(edit-in-editor!)`

A little work to make shebangs possible.
Moves `list-issues` and `new-issue` to a single file, and make it a *pure radicle* file.

In theory `read-file!` no longer has to be a primop. But testing would be quite complicated because of the way files are mocked (which is very relevant to `load!` and so to everything). I'd open a separate issue for that.


exe/RadicleExe.hs
```diff
@@ -76,9 +88,20 @@ opts = Opts
...
+
+argsPrimFn :: forall m . Monad m => [Text] -> PrimFns m
+argsPrimFn args = fromList $ allDocs
+    [ ( "get-args!"
+      , "Returns the list of the command-line arguments the script was called with"
+      , \case
+          [] -> pure . Vec $ fromList (String <$> args)
+          xs -> throwErrorHere $ WrongNumberOfArgs "get-args!" 0 (length xs)
+      )
+    ]
```
geigerzaehler: Can we move this to where all the other primitive functions live? Ideally this should be implemented with capabilities. But we could also have `replBindings` take an argument.

exe/RadicleExe.hs
```diff
@@ -45,6 +49,13 @@ readSource file = do
        then readFile file
        else die $ "Could not find file: " <> toS file
 
+-- | Since `#` is not a comment in radicle, we need to explictly remove shebang
+-- lines.
+ignoreShebang :: Text -> Text
+ignoreShebang src = case T.lines src of
```
geigerzaehler: I think the shebang should be handled gracefully in other cases as well: When reading from stdin, when using `load!` or `read-file!` or `file-module`.

rad/prelude/io-utils.rad
```diff
@@ -12,3 +12,12 @@
   (fn [dict]
     (def result (process-with-stdout! "fzf" [] (string-append (unlines (keys dict)) "\n")))
     (lookup (string-replace "\n" "" (nth 0 result)) dict)))
+
+(def edit-in-editor!
+  "Open `$EDITOR` prepopulated with `orig`. Returns the file saved."
```
geigerzaehler:
```suggestion
  "Open `$EDITOR` on a file prepopulated with `orig`. Returns the contents of the edited file when the editor exits."
```

utils/rad
```diff
@@ -0,0 +1,69 @@
+#!/usr/bin/env radicle
```
geigerzaehler: Can we move this to the `bin` directory? I’d also like to avoid overgeneralization and call the script `rad-issues`. This way we don’t force ourselves yet to build a unified CLI.

exe/RadicleExe.hs
jkarni: It feels to me like `read-file!` should definitely not have it - the file may well not even be radicle code, or code at all. But I can add it to `load!` and `file-module!`.

utils/rad
jkarni: > Can we move this to *the* bin directory

Do you mean to a *new* `bin` directory? There currently isn't one.

Issue #245, opened 2018-11-23
More string functionality


jkarni:
- Regexes
- map
- format
- intercalate
- words/unwords
- lines/unlines
- foldl/foldr

maxhallinan:
Hey @jkarni , is there a reason that `foldr-string` hasn't been implemented? It seem straightforward, which makes me wonder if I'm overlooking something. I'd be happy to add this.

jameshaydon:
@maxhallinan I think there is no reason, please go ahead and add this if you want!

maxhallinan:
Sounds good!

PR #297, opened 2018-12-06
Implement map-string

MeBrei:
Relates to, but does not close, #245.

### DIFF
--- a/rad/prelude/strings.rad
+++ b/rad/prelude/strings.rad
@@ -110,7 +110,7 @@
 
 (def map-string
   "Returns a string consisting of the results of applying `f` to each character
-  of `xs`. Returns a type error if `f` returns something other than a string"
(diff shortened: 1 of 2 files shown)


rad/prelude/strings.rad
```diff
@@ -105,3 +105,21 @@
   [ (lines "foo bar\n\n\nfoo") ==> ["foo bar" "foo"]]
   [ (lines " \n \t") ==> [" " " \t"]]
   [ (lines "") ==> []])
+
+;; map-string
+
+(def map-string
+  "Returns a string consisting of the results of applying `f` to each character
+  of `xs`. Returns a type error if `f` returns something other than a string"
```
jkarni:
```suggestion
  of `xs`. Throws a type error if `f` returns something other than a string"
```

PR #299, opened 2018-12-07
Use localhost for chain tests

geigerzaehler:
Use localhost for chain tests so they can be run locally against `radicle-server`.

### DIFF
--- a/rad/prelude/chain.rad
+++ b/rad/prelude/chain.rad
@@ -44,9 +44,12 @@
     {:chain new-chain
      :result result }))
 
+(def local-test-chain-name
+  (fn [] (string-append "http://localhost:8000/chains/" (uuid!))))
+
 (:test "eval-in-chain"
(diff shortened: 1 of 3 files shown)


rad/prelude/chain.rad
```diff
@@ -44,9 +44,12 @@
     {:chain new-chain
      :result result }))
 
+(def local-test-chain-name
+  (fn [] (string-append "http://localhost:8024/chains/" (uuid!))))
```
jkarni: Isn't the default port the `server` runs on `8000`?

geigerzaehler: I’m using the port that is exposed by `docker-compose.yaml`.

jkarni: Can we make that consistent with the default port? Otherwise there's an extra bit of information that needs to be conveyed for running the tests locally.

PR #301, opened 2018-12-07
Better remote server send! errors

jameshaydon:
The errors returned by a remote server were very hard to read.

### DIFF
--- a/src/Radicle/Internal/Storage.hs
+++ b/src/Radicle/Internal/Storage.hs
@@ -15,6 +15,7 @@ module Radicle.Internal.Storage
 import           Protolude hiding (TypeError)
 
 import           GHC.Exts (fromList)
+import           Servant.Client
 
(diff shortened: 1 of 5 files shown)


src/Radicle/Internal/Storage.hs
```diff
@@ -34,7 +35,7 @@ data StorageBackend m = StorageBackend
 
 -- | Send a list of expressions to a chain identified by the first
 -- argument.
-type StorageSend m = Text -> Seq Value -> m (Either Text ())
+type StorageSend m = Text -> Seq Value -> m (Either ServantError ())
```
geigerzaehler: With this change the `Storage` module becomes dependent on Servant, but the module should be generic. To solve this the error formattaing needs to happen in `HttpStorage`.

jameshaydon: @geigerzaehler done.

PR #310, opened 2018-12-10
Native radicle tests

geigerzaehler:
Currently our testing approach for Radicle code suffers from the following issues

* Mixed approach how tests are written: Partialy as Radicle code with the `:test` macro, partially inlined into haskell.
* Tests running in a mock interpreter in Haskell provide bad debugging experience. For example, you don’t see `print!` outputs.
* Test/debug/fix cycle is slow because you need to recompile haskell

This PR introduces some infrastructure to run tests written with just Radicle code. These tests can now be run with the `./rad/test/run.rad` script.

This requires a mechanism to stub primitive functions which is provided by `rad/tests/stub-primitives.rad`.

**Todo**

- [x] Write documentation on how to write and run Radicle tests
- [x] Integrate `rad/test/run.rad` into CI build.

**Follow-up**

We should create issues for the following follow-up tasks

- [ ] Adjust `rad/monadi/issues.rad` so tests can be run with `radicle rad/test/run.rad`
- [ ] Provide a way to only run a subset of tests.
- [ ] Move tests inlined in Haskell code into Radicle files.
- [ ] Show error messages on test failure.
- [ ] Simplify the way tests can be written. Should just be
      ```
      (:test (def x 42)
             (def y 1)
             (assert-equal (+ x y) 1))
      ```

### DIFF
--- a/docs/source/guide/Testing.rst
+++ b/docs/source/guide/Testing.rst
@@ -1,0 +1,75 @@
+Testing
+=======
+
+Radicle provides a ``:test`` macro which allows you to write tests next
+to your code.
+
+.. code-block:: radicle
+
+  (def not
+    (fn [x] (if x #f #t)))
+
+  (:test "not"
+     [ (not #t) ==> #f ]
+     [ (not #f) ==> #t ]
+     )
+
+The ``test/prelude.rad`` script runs all tests defined in the prelude.
+
+
+Test Definition
+---------------
+
+Each test definition consists of a test name and a list of steps
+
+.. code-block:: radicle
+
+  (:test "my test" step1 step2 ...)
+
+Each step is a vector triple with the symbol ``==>`` in the middle. For
+example
+
+.. code-block:: radicle
+
+  [ (not #t) ==> #f ]
+
+When a test step is run the value left of ``==>`` is evaluated in the
+environment captured at the definition site of the test. The resulting value is
+then compared with the right-hand side. The test passes if both are
+equal. The right-hand side is *not* evaluated.
+
+If the evaluation of the left-hand side throws an error the test fails
+with the message produced by the error.
+
+Changes to reference in a test step evaluation have no effect and all
+tests steps are run independently.
+
+
+Test Setup
+----------
+
+There is a special setup test step that allows you to change the
+environment that tests steps run in.
+
+.. code-block:: radicle
+
+  (:test "with setup"
+    [ :setup (do
+        (def foo 5)
+        )]
+    [ (+ foo 2) ==> 7 ]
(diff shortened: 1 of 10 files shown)


geigerzaehler:
@jameshaydon, @jkarni, @MeBrei: Can I have some feedback on this before I continue the implementation?

jkarni:
I like the general approach. A couple of points/questions:

- It seems to me like we can only stub primfns this way (other things will be defined *after* the stub file). I guess that's okay, but something to consider, and consider whether it's clear to users.
- Now that we have exit codes in radicle, is there any reason to be running the radicle tests from test.hs rather than purely in radicle?

geigerzaehler:
> It seems to me like we can only stub primfns this way (other things will be defined _after_ the stub file). I guess that's okay, but something to consider, and consider whether it's clear to users.

Yes, we can only stub primitives with this approach. Do you have any suggestions for making this clearer to the user?

 

> Now that we have exit codes in radicle, is there any reason to be running the radicle tests from test.hs rather than purely in radicle?

We should be able to port most of the tests to pure radicle. There are a couple of tests that we still want to have in Haskell, though. For example tests for the parser and integration tests between effects and primitive functions.

rad/prelude/test.rad
```diff
@@ -11,6 +17,65 @@
...
+
+(def parse-doc-test-step
+  "Parse a doc test step of the form `[ actual ==> expected ]` and
+  returns a function that when called with an environments runs the
+  step.
+
+  Running a doc step means evluating `actual` in the given environment
+  and comparing the result with `expected` using `assert-equal`."
+  (fn [step]
+    (def lhs (nth 0 step))
```
jameshaydon: Use a `match`? We could make a pattern for matching symbols.

docs/source/guide/Testing.rst
```diff
@@ -0,0 +1,75 @@
+Testing
+=======
+
+Radicle provides a ``:test`` macro which allows you to write test next
```
jameshaydon:
```suggestion
 Radicle provides a ``:test`` macro which allows you to write tests next
```

docs/source/guide/Testing.rst
```diff
@@ -0,0 +1,75 @@
...
+
+  (:test "with setup"
+    [ :setup (do
+        (def foo 5)
+        )]
+    [ (+ foo 2) ==> 7 ]
+    [ (- foo 2) ==> 3 ]
+    )
+
+Similar to test steps the body of the ``:setup`` steps is evaluated in
```
jameshaydon:
```suggestion
 Similar to test steps the body of the ``:setup`` step is evaluated in
```

PR #311, opened 2018-12-10
Remove invalid bytes.

jkarni:
That were causing at least my build/test to choke.


geigerzaehler:
@jkarni: Can you provide more details on why? Rad files should definitely support UTF-8!

jkarni:
> @jkarni: Can you provide more details on why? Rad files should definitely support UTF-8!

I believe it's because the default locale is `C`/`POSIX`, and we don't explicitly state that this is a utf-8-encoded file. While using the system locale for reading most user-created source files seems okay, it seems really cruel to make things then break for people with other locales in the files we distribute when we don't have to. Even if the story is "we *always* assume utf-8" (in which case we should call `setEncoding`), having these quotation marks seems a little gratuitous.

PR #313, opened 2018-12-10
Make a read-file-code! primop, which preserves line-numbers.

jameshaydon:
Creates a primop for reading code from a file. This makes debugging with stack-traces much easier compared to `(read-many (read-file f))` because the resulting values are annotated with line-numbers from `f` (at least when running chains locally, errors from remote chains are still obscure).


src/Radicle/Internal/PrimFns.hs
```diff
@@ -75,24 +75,26 @@ purePrimFns = fromList $ allDocs $
...
           [fn, List args] -> callFn fn args
           [_, v]          -> throwErrorHere $ TypeError "apply" 1 TList v
           xs -> throwErrorHere $ WrongNumberOfArgs "apply" 2 (length xs))
-    , ( "read"
+    , ( "read-annotated"
       , "Parses a string into a radicle value. Does not evaluate the value."
-      , oneArg "read" $ \case
-          String s -> readValue s
-          v -> throwErrorHere $ TypeError "read" 0 TString v
+      , twoArg "read-annotated" $ \case
```
geigerzaehler: Could you update the doc string so that it mentions the second argument? Same for `read-many-annotated`.

src/Radicle/Internal/PrimFns.hs
```diff
@@ -668,8 +671,8 @@ readValues
     => Text
     -> Text
     -> m [Value]
-readValues name s = do
-    let p = parseValues ("[" <> name <> "-primop]") s
+readValues label s = do
+    let p = parseValues ("[" <> label <> "]") s
```
geigerzaehler: This conflicts with #300. I suggest to wrap the label in square brackets at the call site, i.e. the definition of the primitives.

rad/prelude/io.rad
```diff
@@ -1,25 +1,29 @@
...
+  (fn [] (read-annotated "read-line!" (get-line!))))
 
-(def read-code!
-  "Read code (as data) from a file. Returns a vector of expressions"
-  (fn [filename]
-    (read-many (read-file! filename))))
+(def read-file-value!
+  "Read a single radicle value from a file."
+  (fn [file]
+    (read-annotated file (read-file! file))))
```
geigerzaehler: This will wrap the filename in square brackets which is in contrast to `load!`. Is this what we want?

PR #314, opened 2018-12-10
Add end-to-end tests for radicle-server to CI

geigerzaehler:
We add a CI build step that runs the tests for the counter example app against a build of `radicle-server`.


geigerzaehler:
I added an extra commit that proofs that we fail if the server is broken. I’ll remove that before merging

Issue #305, opened 2018-12-10
Adding comments to issues via the CLI


jkarni:
After an issue is opened in the editor, there should be a way to add a comment.

PR #315, opened 2018-12-11
Implement take-while for sequences

MeBrei:
This while help solving #305

### DIFF
--- a/rad/prelude/list.rad
+++ b/rad/prelude/list.rad
@@ -81,14 +81,16 @@
 ;; take-while
 
 (def-rec take-while
-  "Returns all elements of a list `ls` until one does not satisfy `pred`"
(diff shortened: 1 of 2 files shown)


rad/prelude/list.rad
```diff
@@ -74,3 +75,18 @@
   [ (filter (fn [x] (< x 10)) (list 3 10 11))
      ==> (3) ]
 )
+
+;; take-while
+
+(def-rec take-while
+  "Returns all elements of a list `ls` until one does not satisfy `pred`"
```
jkarni: After #277 this should work on vectors as well. Might be worth waiting on it, and then changing the docs here to say `sequence`, and the tests to include vectors.

PR #316, opened 2018-12-11
Parametrized issues chain

jkarni:
Parametrizes issues functions by chain. Adds the chain name to the CLI as an option. Also:

- Makes `take` work with strings
- Makes `issues.rad` not load the chain automatically on import/load.

### DIFF
--- a/bin/rad-issues
+++ b/bin/rad-issues
@@ -2,7 +2,10 @@
 
 (load! "rad/prelude.rad")
 
+(def base-name "http://machine.radicle.xyz/chains")
+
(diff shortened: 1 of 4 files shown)


bin/rad-issues
```diff
@@ -53,19 +70,25 @@
 ")
     (def iss (read (edit-in-editor! template)))
     (create-issue!
-         (insert
-           :state :open
+      chain-ref
+      (insert
```
jameshaydon: Use `<>`, or wait for #321.

PR #318, opened 2018-12-11
More functions in `prelude/set`.

jameshaydon:


rad/prelude/set.rad
```diff
@@ -17,14 +21,31 @@
     (delete x s)))
 
 (def member?
-  "Query if an value is an element of a set."
-  (fn [x s]
-    (member? x s)))
+  "Query if a value is an element of a set."
```
jkarni: Looks like the fn def was lost?

rad/prelude/set.rad
```diff
@@ -17,14 +21,31 @@
...
 
 (def to-vec
   "Convert a set to a vector."
-  (fn [s] (keys s)))
+  (fn [s] (dict/keys s)))
 
 (def from-seq
   "Create a set from a sequence."
-  (fn [xs] (foldr set/insert {} xs)))
+  (fn [xs] (foldr insert {} xs)))
```
jkarni: Maybe a test here (e.g., to check `(from-seq [1 2])` == `(from-seq [1 2 2])`))

rad/prelude/set.rad
jameshaydon: Sets are implemented as dicts so `set/member?` is just an alias for `member?`.

jkarni: Hm, somehow didn't see the next line.

rad/prelude/set.rad
```diff
@@ -17,14 +21,35 @@
...
+(def key-set
+  "The set of keys of a dict."
+  (fn [d]
+    (map-values (fn [v] #t) d)))
+
+(:test "key-set"
+  [ (key-set {:a 1 :b 2}) ==> {:a #t :b #t} ])
+
+(def subset?
+  "Checks if xs is a subset of ys."
```
MeBrei: Maybe add backticks around the variables for better visualization of the input.

jameshaydon:
@jkarni @MeBrei bump

PR #319, opened 2018-12-11
Better validation of dicts

jameshaydon:
Decoupling of
- validating existence of keys,
- validating values associated to keys.

- [ ] tests


rad/prelude/validation.rad
```diff
@@ -58,18 +60,43 @@
...
+      (if (member? k x)
+        x
+        (throw 'validation-failure {:info "Did not contain key."
+                                    :element k
+                                    :container x})))))
+
+(def has-key
+  "Combines existence and validity of a key in a dict."
+  (fn [k v]
+    (and [(contains k) (key k v)])))
```
geigerzaehler: What do you think about calling this `key` and renaming `key` to `optional-key`? I suspect devs are going to gravitate towards `key` when they actually want `has-key`. I also feel the documentation is a bit to sparse. Can we use the original documentation for `key`?

rad/prelude/validation.rad
```diff
@@ -79,6 +106,27 @@
...
+  (fn [ks]
+    (fn [d]
+      (if (set/subset? (set/key-set d) (set/from-seq ks))
+        d
+        (throw 'validation-failure {:info "Dict contained illegal keys."
+                                    :dict d
+                                    :allowed-keys ks})))))
+
+(def has-keys
+  "Checks for the existence and validity of keys."
```
geigerzaehler: Could you update the doc strings or argument names so that the generated documentation looks nice?

PR #320, opened 2018-12-11
[RFC] Machine Backend Interface

geigerzaehler:
[Rendered preview](https://github.com/oscoin/radicle/blob/rfc/machine-backend-interface/rfcs/0001-machine-backend-interface.rst)

**Synopsis**

We propose an interface for backends that control Radicle State Machines. This interface replaces the `StorageBackend`` interface defined in ``Radicle.Internal.Storage``. It’s design is motivated by the requirements of implementing the interface for IPFS.

### DIFF
--- a/rfcs/0001-machine-backend-interface.rst
+++ b/rfcs/0001-machine-backend-interface.rst
@@ -1,0 +1,72 @@
+Machine Backend Interface Definition
+====================================
+
+.. date:: 2018-12-11
+.. reviewers::
+
+   @jkarni
+   @jameshaydon
+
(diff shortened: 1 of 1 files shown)


geigerzaehler:
> I'm just wondering if we can get rid of the `machineType` and just infer this from the `MachineId`? e.g. `ipfs://` vs `http://`?

My motivation for `machineType` was to generate the primitive function names. (I forgot to mention that in the document.) I’ll likely omit it in the first implementation and we can have some more discussions on URLs and machine IDs.

PR #321, opened 2018-12-11
Implement adding comments to list-issues

MeBrei:
When viewing an issues via 'list-issues', a comment can be entered at the top.

### DIFF
--- a/bin/rad-issues
+++ b/bin/rad-issues
@@ -71,13 +71,7 @@
 ")
     (def iss (read (edit-in-editor! template)))
     (create-issue!
-         (<>
-           {:state :open}
-           (<>
-              {:comments []}
-              (<>
-                {:created-at (now!)}
-                  iss))))))
+         (<> iss {:state :open :comments [] :created-at (now!)}))))
 
 (def args (get-args!))
 (match args
--- a/bin/rad-issues
+++ b/bin/rad-issues
@@ -21,9 +21,8 @@
     (load! "rad/monadic/issues.rad")
 
     (def mk-key (fn [v]
-      (def key (nth 0 v))
-      (def val (nth 1 v))
-      [(string-append (show key) " - " (lookup :title val)) val]))
+      (match v
+        ['key 'val] [(string-append (show key) " - " (lookup :title val)) val])))
     (def iss (dict-from-seq (map mk-key (seq (list-issues)))))
     (def selected (fzf-select-by-key! iss))
     (def comment-divider "---")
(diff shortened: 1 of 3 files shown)


jameshaydon:
Some comments, not related to the current diff:

`(load! "rad/monadic/issues.rad")`
Maybe we should make the imports of `issues.rad` a little leaner so that this CLI tool is a little more performant.

```
(insert
           :state :open
           (insert
              :comments []
              (insert
                :created-at (now!)
iss)))
```

One can use `<>`.

```
(def key (nth 0 v))
(def val (nth 1 v))
```
Use `match`.

```
(dict-from-seq (map mk-key (seq (list-issues))))
```
I think `map-keys` can be used here.

bin/rad-issues
```diff
@@ -26,19 +26,38 @@
       [(string-append (show key) " - " (lookup :title val)) val]))
     (def iss (dict-from-seq (map mk-key (seq (list-issues)))))
     (def selected (fzf-select-by-key! iss))
+    (def comment-divider "---")
+    (def username
+      (fn [d]
+        (catch 'any (lookup :github-username d) (fn [x] "???"))))
```
jameshaydon: I think we have `@def` for this.

bin/rad-issues
```diff
@@ -26,19 +26,38 @@
...
+    (def issue-number (fn [d] (lookup :number d)))
+    (def body (fn [d] (lookup :body d)))
     (def pretty-comment (fn [c]
       (string-append
-        "* **" (lookup :github-username c) "** [" (lookup :created-at c) "]:\n\n"
-        (string/unlines (map (fn [xs] (string-append "   " xs)) (string/lines (lookup :body c)))))))
+        "* **" (username c) "** [" (lookup :created-at c) "]:\n\n"
+        (string/intercalate "\n\n" (map (fn [xs] (string-append "   " xs)) (string/lines (body c)))))))
```
jameshaydon: Why the empty lines? I.e., why not use `string/unlines`?

bin/rad-issues
```diff
@@ -26,19 +26,38 @@
...
+         "\n## Comments\n\n\n" (string/unlines (map pretty-comment (lookup :comments i)))
     ))))
-    (print! (edit-in-editor! (pretty-issue selected)))))
+    (def edited-issue (edit-in-editor! (pretty-issue selected)))
+    (def extract-comment
+      (fn [xs]
+        (string/unlines (take-while (fn [x] (not (eq? x comment-divider))) xs))))
+    (def send-comment
+      (fn [xs]
+        (def comment (extract-comment (string/lines edited-issue)))
```
jameshaydon: I might move the `string/lines` into `extract-comment`.

bin/rad-issues
MeBrei: `string/unlines` adds `\n` between the lines, this was not detected as a linebreak by the markdown preview though.

bin/rad-issues
jameshaydon: Actually, it's only added in #324.

bin/rad-issues
jameshaydon: Ah I see.

PR #326, opened 2018-12-11
Add logging to `radicle-server`

geigerzaehler:
Add logging to `radicle-server`


exe/Server.hs
```diff
@@ -124,6 +132,7 @@ loadState conn = do
 -- guarantee).
 submit :: Connection -> Chains -> Text -> Values -> Handler Value
 submit conn chains name (Values vals) = do
+    logInfo "Submit expressions" [("machine", name), ("expression-count", show $ length vals)]
```
jkarni:
```suggestion
    logInfo "Submitted expressions" [("machine", name), ("expression-count", show $ length vals)]
```

PR #334, opened 2018-12-12
Generalization of the storage interface

geigerzaehler:
As a step towards RFC-0001 we implement a couple of changes. Updates to both the interpreter and the server are backwards incompatible.

* `StorageBackend` is now polymorphic in the entry index and uses the `ToRad` and `FromRad` instances.
* `storageSend` now returns the index of the submitted expression. This also requires an update of `insertExprs` in the server code.
* The index parameter to `storageReceive` is now optional.
* `storageReceive` returns an index for the last entry.
* Change the semantics of `storageReceive` so that we don’t include the item at the given index.
* Update `prelude/chain` to accommodate the interface changes.
* Add `test/server.rad` script that tests the interface against the HTTP
  server.

### DIFF
--- a/src/Radicle/Internal/HttpStorage.hs
+++ b/src/Radicle/Internal/HttpStorage.hs
@@ -59,15 +59,17 @@ instance MimeUnrender PlainText Values where
               where err' :: LangError Value
                     err' = LangError [thisPos] (ParseError err)
 
--- | Endpoint to submit an expression to a chain. Responds with @:ok@ if the
--- expression has been submitted sucessfully.
+-- | Endpoint to submit an expression to a chain. Responds with the
+-- index of the submitted expression. The index is an integer encoded
+-- as a Radicle value.
 type ChainSubmitEndpoint = "submit" :> ReqBody '[PlainText] Values :> Post '[PlainText] Value
 
 chainSubmitEndpoint :: Proxy ChainSubmitEndpoint
 chainSubmitEndpoint = Proxy
 
--- | Endpoint to obtain all entries in a chain starting from the given
--- index. Responds with a Radicle vector of the expressions.
+-- | Endpoint to obtain all entries in a chain starting from and
+-- including the given index. Responds with a Radicle vector of the
+-- expressions.
 type ChainSinceEndpoint = "since" :> Capture "index" Int :> Get '[PlainText] Values
 
 chainSinceEndpoint :: Proxy ChainSinceEndpoint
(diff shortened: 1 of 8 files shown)


src/Radicle/Internal/HttpStorage.hs
```diff
@@ -94,15 +94,22 @@ httpStoragePrimFns' mgr =
               \ chain located at the URL for evaluation."
```
jameshaydon: Add documentation on the return value.

src/Radicle/Internal/HttpStorage.hs
```diff
@@ -94,15 +94,22 @@ httpStoragePrimFns' mgr =
...
-                pure $ bimap formatServantError (const ()) res
+                pure $ case res of
+                    Left servantError -> Left $ formatServantError servantError
+                    Right val -> case fromRad val of
+                        Left err -> Left $ "cannot parse server response: " <> err
+                        Right n -> Right n
             )
         , storageReceive =
             ( "receive!"
             , "Given a URL (string) and a integral number `n`, queries the remote chain\
```
jameshaydon: Add note that it is now a `Maybe Number`.

jameshaydon:
So I think the `fromRad`/`toRad` is a bit confusing in the Http API.

Here:
https://github.com/oscoin/radicle/blob/f1b03ec228e5ce8e2d1cb9609e44569ab8770796/src/Radicle/Internal/HttpStorage.hs#L62

```haskell
-- | Endpoint to submit an expression to a chain. Responds with @:ok@ if the
-- expression has been submitted sucessfully.
type ChainSubmitEndpoint = "submit" :> ReqBody '[PlainText] Values :> Post '[PlainText] Value
```

we should update the docs since the `:ok` is not a number. But also, why not just make it and `Int` rather than a `Value`?

src/Radicle/Internal/HttpStorage.hs
```diff
@@ -94,15 +94,22 @@ httpStoragePrimFns' mgr =
...
               \ for the last `n` inputs that have been evaluated."
-            , \url index -> do
+            , \url maybeIndex -> do
+                let index = fromMaybe (-1) maybeIndex
                 res <- liftIO $ runClientM' url mgr (since index)
-                pure $ first formatServantError res
+                pure $ case res of
+                    Left err     -> Left $ formatServantError err
+                    Right values -> Right (length values + index, values)
```
jameshaydon: The `length value + index` is depending on implementation details of the Http server. Maybe it would be best to make the Http server's API more similar to a `StorageBackend`, that is, return the index of the last input returned.

geigerzaehler:
@jameshaydon: I fixed all the documentation. I also reconsidered the approach towards the server. I reverted most of the changes to the server code to keep the HTTP API backwards compatible. We do all the required adjustments on the client side now. I’ll follow up with a PR that consolidates the HTTP API and the storage interface.

PR #335, opened 2018-12-12
Name chain

jkarni:
Hurriedly-written name chain. @MeBrei and I are going to start working on using these names now.

### DIFF
--- a/rad/monadic/names-remote.rad
+++ b/rad/monadic/names-remote.rad
@@ -33,7 +33,6 @@
   (fn []
     (read-ref names)))
 
-(def t (fn [] "hi"))
 
 ;; Commands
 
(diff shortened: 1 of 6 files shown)


rad/monadic/names-remote.rad
```diff
@@ -0,0 +1,75 @@
+(import prelude/basic :unqualified)
+(import prelude/bool :unqualified)
+(import prelude/dict :unqualified)
+(import prelude/ref :unqualified)
+(import prelude/validation :as 'validator)
+(import prelude/patterns :unqualified)
+(import prelude/util :unqualified)
+(import prelude/lens :unqualified)
+
+(def names (ref []))
```
jameshaydon: any reason for not using `{}`?

jkarni: It feels like we want these to be look-upable and unique both ways, and between having two dicts or one list of pairs, it seemed like a toss.

jameshaydon: :+1:

PR #336, opened 2018-12-12
Use a custom stack snapshot and improve CI cache

geigerzaehler:
We now use a custom stack snapshot instead of specifying `extra-deps` in `stack.yaml` and tweak the CI caching scripts.

Before the CI build caching suffered from a couple of problems:

* A cache for `STACK_ROOT` would always be uploaded after the build. (I don’t know why.)
* We cached the build artifacts in `.stack-work` keyed by the has of `stack.yaml`. This resulted in many builds sharing building artifacts which could potentially result in bugs.

With this commit we stop caching build artifacts and improve the caching strategy for `STACK_ROOT`.

By using a snapshot the artifacts for the extra packages (e.g. `cbor`) are build in `STACK_ROOT` and cached. (Including the extra packages in the cache was the initial reason for caching `.stack-work`)

The main cache ID for `STACK_ROOT` is determined by the hash of `snapshot.yaml` and `package.yaml`. (Ideally the cache ID should be the hash of the output of `stack ls dependencies`, but this requires some further modifications.)

In addition we have a `STACK_ROOT` cache for the master branch. This cache is always uploaded for a master build. If a cache with a specific hash does not exist we load the master cache.

### DIFF
--- a/cloudbuild.yaml
+++ b/cloudbuild.yaml
@@ -1,4 +1,4 @@
-timeout: 1800s # Cache misses are slow to rebuild
+timeout: 3600s # Cache misses are slow to rebuild
 steps:
   - id: "Load cache"
     name: gcr.io/cloud-builders/gsutil
@@ -23,7 +23,7 @@ steps:
       stack config set system-ghc --global true
       stack config set install-ghc --global false
 
-      stack install stylish-haskell hlint weeder
+      stack build stylish-haskell hlint weeder
 
   - id: "Format"
     name: 'haskell:8.4.3'
@@ -150,6 +150,8 @@ steps:
     waitFor:
     - "Build & Test"
     name: gcr.io/cloud-builders/gsutil
+    env:
+    - BRANCH_NAME=$BRANCH_NAME
     entrypoint: 'bash'
     args:
     - '-c'
--- a/scripts/ci.sh
+++ b/scripts/ci.sh
@@ -2,30 +2,40 @@
 
 set -euo pipefail
 
+dependency_hash=$( (cat package.yaml; cat snapshot.yaml) | sha256sum | cut -d ' ' -f 1)
+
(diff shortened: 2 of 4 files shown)


cloudbuild.yaml
```diff
@@ -23,7 +23,7 @@ steps:
       stack config set system-ghc --global true
       stack config set install-ghc --global false
 
-      stack install stylish-haskell hlint weeder
+      stack build stylish-haskell hlint weeder
```
kim: This still irks me. We should really maintain our own build image with these tools baked in.

scripts/ci.sh
```diff
@@ -2,30 +2,41 @@
 
 set -euo pipefail
 
+dependency_hash=$( (cat package.yaml; cat snapshot.yaml) | sha256sum | cut -d ' ' -f 1)
+
+remote="gs://radicle-build-cache/v2"
+
+local_cache_archive="stack-root.tar.gz"
+remote_cache_master="$remote/stack-root-master.tar.gz"
```
kim: I’m not sure I understand the reason for two cache bundles - why is master different?

geigerzaehler: We have master caches to make changes to the dependencies still be fast: Assume we add a dependency and build the code for the first time. No cache exists for the calculated key. Without the master cache we would rebuild the code from scratch. With the master cache we only need to build the added dependencies.

I made some changes to the `save-cache` function so that the overhead adding that cache should be minimal. We just copy the existing hashed cache in the cloud.

geigerzaehler:
> Do snapshots support git dependencies? I figure we will spend a lot of time building those redundantly down the road.

Yes, they support git dependencies. Fetching the source is done locally in `.stack-work` so it needs to be repeated for every build. But then the source is hashed in the build happens in the global stack root which is cached. So on a subsequent build we check out the source again but then find the already built version.

PR #342, opened 2018-12-13
Improve CLI parsing and error msgs for rad-issues

geigerzaehler:
Improve CLI parsing and error messages for rad-issues


geigerzaehler:
@MeBrei Addressed your comment

PR #345, opened 2018-12-14
Primfns for local storage at an XDG file.

jameshaydon:
Adds Primfs for reading/writing to persistent storage (currently implemented as an XDG file).

- [x] Tests


src/Radicle/Internal/Effects.hs
```diff
@@ -337,6 +337,19 @@ replPrimFns sysArgs = fromList $ allDocs $
...
           v -> throwErrorHere $ TypeError "file-module!" 0 TString v
       )
+    , ( "read-persistent-state!"
+      , "Reads the persistent state."
+      , \case
+          [] -> do v_ <- readPersistentState
+                   case v_ of
+                     Left e  -> throwErrorHere $ OtherError e
+                     Right v -> pure v
+          xs -> throwErrorHere $ WrongNumberOfArgs "read-persistent-state!" 0 (length xs)
```
geigerzaehler: Couldn’t we implement this in pure radicle using `read-file!` and `write-file!`? Then we would have to possibilities for the filename: Either expose `getLocalStateFile` as a primitve or don’t use XDG for now. I’d lean towards the later to keep the number of primitives low.

jameshaydon: If we don't use XDG, which file location do we use?

geigerzaehler: It’s up to the app. So for example `rad-issue` would use ~/.rad-issue.db.rad` or something.

jameshaydon: Hm, but I kind of like the idea of a *global* local storage. Some keys might be used by several apps, some apps might want to read keys of other apps they know how to work with, etc.

geigerzaehler: They still can co-operate on an ad-hoc basis. I think there is a danger to every app sharing the same database by default.

PR #350, opened 2018-12-18
Server loads chains into memory on demand

geigerzaehler:
The server now loads chains into memory on demand instead of all chains at once. This eliminates a bug where we did not get all the expressions because of the default row limit.


.hlint.yaml
```diff
@@ -31,4 +31,4 @@
 - ignore: {name: Eta reduce}
 - ignore: {name: Use lambda-case}
 - ignore: {name: Redundant $}
-- ignore: {name: Redundant do}
```
geigerzaehler: “Redundant do” was redundant. It is listed at the top of the list

jkarni:
Note that this means we won't know, when doing a server upgrade, whether
the new server can handle the existing transactions (e.g. no primfn
renamed), which I think was quite nice (k8s would not do such incompatible
deploys) and it'd be sad to see it go.

(I may be wrong here - on my phone and it read the PR description)

Am Di., 18. Dez. 2018, 11:40 hat Thomas Scholtes <notifications@github.com>
geschrieben:

kim:
Drive by: are you assuming all transactions fit in memory? `postgresql-simple` has ways to [airquotes-stream-airquotes](https://hackage.haskell.org/package/postgresql-simple-0.6/docs/Database-PostgreSQL-Simple.html#v:fold) query results.

geigerzaehler:
> Drive by: are you assuming all transactions fit in memory? `postgresql-simple` has ways to [airquotes-stream-airquotes](https://hackage.haskell.org/package/postgresql-simple-0.6/docs/Database-PostgreSQL-Simple.html#v:fold) query results.

The size of the `Chain` object is linear in the number of inputs which is the number of rows in the result. So for now we have to assume that the result set fits into memory. Nevertheless we should use the `fold` to be independent of the default row limit.

> Note that this means we won't know, when doing a server upgrade, whether the new server can handle the existing transactions (e.g. no primfn renamed), which I think was quite nice (k8s would not do such incompatible deploys) and it'd be sad to see it go. (I may be wrong here - on my phone and it read the PR description)

You’re correct in you parsing. We will miss this feature. Instead we will see the errors when we try to update a machine. On the other hand the current approach does not scale and we might actually want to be incompatible with broken chains.

jkarni:
> You’re correct in you parsing. We will miss this feature. Instead we will see the errors when we try to update a machine. On the other hand the current approach does not scale and we might actually want to be incompatible with broken chains.

> On the other hand the current approach does not scale

That may be true, but we can always just load chains one by one (and in batches if needed) on startup if that's the problem, and still fail early cancelling the deployment if the upgrade is incompatible. 

> we might actually want to be incompatible with broken chains.

Possibly, but we should have to be explicit about this. Right now things are just going to break, and we won't know until we try using our issues chain. And if we happen to create a new chain before that, which uses new features... well, then there isn't even any rollback to a state that can deal with all the data.

PR #351, opened 2018-12-18
Correctly order results when loading initial state

geigerzaehler:
Smaller fix for problem addressed in #350

This fix an issue where the server failed to load the machines from the database. The underlying problem was that rows from the DB where not correctly grouped by chain name. The `groupBy` function only groups *adjacent* items in a list.

### DIFF
--- a/exe/Server/DB.hs
+++ b/exe/Server/DB.hs
@@ -25,11 +25,14 @@ getSinceDB conn name id
 -- | Get all txs in all chains. Useful after a server restart.
 getAllDB :: Connection -> IO [(Text, [Value])]
 getAllDB conn = do
-    res :: [(Int, Text, Value)] <- query_ conn "SELECT id, chain, expr FROM txs"
+    -- The @ORDER BY@ clauses are important, because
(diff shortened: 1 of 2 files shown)


exe/Server/DB.hs
```diff
@@ -25,11 +25,14 @@ getSinceDB conn name id
 -- | Get all txs in all chains. Useful after a server restart.
 getAllDB :: Connection -> IO [(Text, [Value])]
 getAllDB conn = do
-    res :: [(Int, Text, Value)] <- query_ conn "SELECT id, chain, expr FROM txs"
+    -- The @ORDER BY@ clauses are important, because
+    -- * 'groupBy' only groups adjacent items in a list, and
```
jameshaydon: Maybe use a `GROUP BY`?

geigerzaehler: As far as I understand it `GROUP BY` in SQL only makes sense for aggregates. How would you use it to make the groups here?

jameshaydon: Ah yeah I got a bit to used to neo4j. You can use `array_agg` but probably not worth it.

PR #352, opened 2018-12-18
[NOMERGE] Machine environments and states now have their own value types.

jameshaydon:
**Results in about 3.6 times faster machines! :rocket:**

This PR adds two value types:
- _environments_: i.e. what gets captured in a closure or module,
- _states_: represents the state of a (pure) radicle machine.

This should make `eval`s faster (becasue it removes all the back-and-forth translations) and less error-prone (because manipulations use special-purpose functions).

Furthermore:
- Adds some primitive functions for manipulating these values.
- Renames some of the primfns (the code used `env` to refer to the whole state a lot, which is misleading).
- The issues chain is broken by these changes, so an upgrade script is included. At the same time we make the issues chain fix missing `:modified-at` keys on the fly.
- Lists are treated as patterns (same as vectors).
- Add the `/member` pattern for matching against several possible constants.

### DIFF
--- a/src/Radicle/Internal/Core.hs
+++ b/src/Radicle/Internal/Core.hs
@@ -5,12 +5,12 @@
 module Radicle.Internal.Core where
 
 import qualified Prelude
-import           Protolude hiding (Constructor, Handle, TypeError, (<>))
+import           Protolude hiding (Constructor, Handle, State, TypeError, (<>))
 
 import           Codec.Serialise (Serialise)
 import           Control.Monad.Except
                  (ExceptT(..), MonadError, runExceptT, throwError)
-import           Control.Monad.State
+import           Control.Monad.State hiding (State)
 import           Data.Aeson (FromJSON(..), ToJSON(..))
 import qualified Data.Aeson as A
 import           Data.Copointed (Copointed(..))
@@ -25,7 +25,6 @@ import           Generics.Eot
 import qualified GHC.Exts as GhcExts
 import qualified GHC.IO.Handle as Handle
 import           System.Process
-                 (CmdSpec(..), CreateProcess(..), ProcessHandle, StdStream(..))
 import qualified Text.Megaparsec.Error as Par
 
 import           Radicle.Internal.Annotation (Annotated)
@@ -255,6 +254,8 @@ data ValueF r =
     -- The first argument is the name for recursive calls to the
     -- function in the body.
     | LambdaRecF Ident [Ident] (NonEmpty r) (Env r)
+    | VEnvF (Env r)
+    | VStateF State
     deriving (Eq, Ord, Read, Show, Generic, Functor)
 
 instance Serialise r => Serialise (ValueF r)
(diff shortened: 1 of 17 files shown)


src/Radicle/Internal/Core.hs
```diff
@@ -440,6 +455,13 @@ instance GhcExts.IsList (PrimFns m) where
     fromList xs = PrimFns . Map.fromList $ [ (i, Doc.Docd d x)| (i, d, x) <- xs ]
     toList e = [ (i, d, x) | (i, Doc.Docd d x) <- GhcExts.toList . getPrimFns $ e]
 
+data PureState = PureState
+  { env  :: Env Value
+  , refs :: IntMap Value
+  } deriving (Eq, Ord, Read, Show, Generic)
```
geigerzaehler: Can we find a better name than `PureState`? Maybe just `State` or `EvaluatorState`? What about prefixing the fields to avoid name clashes?

src/Radicle/Internal/PrimFns.hs
```diff
@@ -69,12 +70,41 @@ purePrimFns = fromList $ allDocs $
...
-          xs -> throwErrorHere $ WrongNumberOfArgs "pure-env" 0 (length xs)
+          xs -> throwErrorHere $ WrongNumberOfArgs "pure-state" 0 (length xs)
+      )
+    , ( "state->env"
+      , "Extract the environment from a radicle state."
+      , oneArg "state->env" $ \case
+          State s -> pure (VEnv (env s))
+          v -> throwErrorHere $ TypeError "state->env" 0 TState v
+      )
+    , ( "set-binding"
```
geigerzaehler: How about making `set-binding` and `get-binding` operate on a state rather than the environment? This would eliminate the need for the `VEnvF` value type and the `state->env` and `set-env` functions and would make it generally nicer to use.

geigerzaehler: We are also inconsitent with names. We use the term `binding` here for what we call environment in the Haskell codebase. There `binding` refers to more stuff like the primitives and handles.

rad/prelude/test.rad
```diff
@@ -27,6 +27,7 @@
   ```
   "
   (fn [test-def]
+    (put-str! (show test-def))
```
geigerzaehler: Oops

src/Radicle/Internal/PrimFns.hs
jameshaydon: > This would eliminate the need for the VEnvF value type

Yes, I thought about this, but there is one place where environments are used standalone: modules. Currently a module is dict:

```clojure
{:module 'foo
 :export ['bar]
 :env <env>}
```

I agree that the naming is inconsistent in the codebase (especially the radicle code which refers to the full-state as `env` a lot). I think the most important is to present clear naming on the radicle side (which is what we expose to users), and this shouldn't be affected by the haskell source (which we can update later). I suggest:

- `state` for the machine-state,
- `env` for the enviornment, part of the `state`,
- a `binding` refers to a particular `(identifier, value)` pair an an `env`.

geigerzaehler: I like the terminology.

Too bad we can’t get rid of `VEnvF`. Maybe we should consider using state for modules in the future.

jameshaydon: The part of the `state` that is not `env` is the state of the references; I don't think that should be captured as part of a module definition.

PR #353, opened 2018-12-18
Fix bad rebase introduced in c917a20

geigerzaehler:


bin/rad-issues
```diff
@@ -219,7 +218,7 @@
 
 (def args (get-args!))
 (match args
-  (/cmd-1 "list" 'chain-name) (list-issues chain-name)
+  (/cmd-1 "list" 'chain-name) (list-items chain-name :nothing)
   (/cmd-1 "list-new" 'chain-name) (list-items chain (read-file-key! ".radicle-issues" :last-mark-done))
```
MeBrei: `chain` here also has to be replaced by `chain-name`

PR #356, opened 2018-12-19
RFC: Fast, incremental app state updates

jameshaydon:
This RFC proposes full serialisation of the radicle interpreter state as a
solution to maintaining up-to-date views/projections of the state of a
long-running machine that is relevant to an app.

[render](https://github.com/oscoin/radicle/blob/rfc/serialise-state/rfcs/0002-machine-state-catchup.rst)

### DIFF
--- a/rfcs/0002-machine-state-catchup.rst
+++ b/rfcs/0002-machine-state-catchup.rst
@@ -1,5 +1,5 @@
-Fast, incremental radicle app state
-===================================
+Fast radicle app state updates
+==============================
 
 .. date:: 2018-12-19
 .. reviewers::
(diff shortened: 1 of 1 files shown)


jameshaydon:
@cloudhead 

> choice of backend

Not entirely sure what you mean by "backend", are you referring to the representation of the interpreter state (stage 1), or a choice of DB (stage 2)?

rfcs/0002-machine-state-catchup.rst
```diff
@@ -0,0 +1,173 @@
...
+
+In order to be able to serialise the state despite (1) and (2), the interpreter
+implementation is changed to accommodate a new representation of environments:
+
+- Environments are maps from identifiers to *hashes* of values.
+
+- The hashes are treated as pointers into a global values map, from hashes to
+  values.
+
+- Environments are hierarchical, so that the sharing is explicit.
```
geigerzaehler: Could you expand on this a bit? What does “hierarchical” mean?

rfcs/0002-machine-state-catchup.rst
```diff
@@ -0,0 +1,173 @@
...
+  reasons.
+
+We propose two stages:
+
+.. _stage1:
+
+**Stage 1: recreating radicle state**
+
+In order to be able to serialise the state despite (1) and (2), the interpreter
+implementation is changed to accommodate a new representation of environments:
```
geigerzaehler: I’d love to see more details on this. Almost a sketch of the implementation.

rfcs/0002-machine-state-catchup.rst
```diff
@@ -0,0 +1,173 @@
...
+solution to maintaining up-to-date views/projections of the state of a
+long-running machine that is relevant to an app. For facilitating deriving these
+projections, it proposes that machine outputs are write-commands for a
+particular database.
+
+Motivation
+----------
+
+An *app* (e.g. the radicle issues CLI) is a program which reads data from a
+radicle machine, displays it to the user, and facilitates ``send!``ing inputs to
```
geigerzaehler: RST does not support this :(. You need to leave a space between the closing <code>``</code> and “ing”.

cloudhead:
> Not entirely sure what you mean by "backend", are you referring to the representation of the interpreter state (stage 1), or a choice of DB (stage 2)?

I was referring more to the DB, since in memory you can have whatever representation you want. It's also possible that we decide to use a different representation of both 1. and 2. (or maybe just 2.) when in a blockchain context. That's also an option.

rfcs/0002-machine-state-catchup.rst
jameshaydon: :sob:

jameshaydon:
@cloudhead 

The idea for a "DB" I had most in mind was just one coded in radicle itself, backed by a triple-store (a set of vectors of length 3), so I don't think this would change much in terms of how we query for "keys" (I would imagine that the first to elements of the triple would be considered a key).

However, I think maybe we should start a separate discussion on how key lookups are going to work.

cloudhead:
@jameshaydon I see - I was thinking of a design where the state isn't read in memory, but queried from an on-disk database. I think this will eventually be necessary as the state grows larger, but for now it seems like it's off-topic.

geigerzaehler:
@jameshaydon Shall we merge this? Although we’ll not work on it in the near future I think it is a nice thing to have in the docs.

jameshaydon:
@geigerzaehler :+1: I will add some notes about cyclicity being resolved and some of the questions that came up during experimentation and then merge.

PR #360, opened 2018-12-20
Some improvements to primFns

MeBrei:
* consistently use the `tt` `ff` variables

* unify alignments of some functions

* add string functionality to `drop`

* simplify code for `/`

### DIFF
--- a/src/Radicle/Internal/PrimFns.hs
+++ b/src/Radicle/Internal/PrimFns.hs
@@ -375,8 +375,8 @@ purePrimFns = fromList $ allDocs $
       , "Checks if a number is an integer."
       , oneArg "integral?" $ \case
           Number n -> case Num.isInteger n of
-            Left _  -> pure $ Boolean False
-            Right _ -> pure $ Boolean True
+            Left _  -> pure ff
+            Right _ -> pure tt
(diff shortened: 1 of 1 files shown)


src/Radicle/Internal/PrimFns.hs
```diff
@@ -351,8 +352,8 @@ purePrimFns = fromList $ allDocs $
     , ( "/"
       , "Divides one number by another. Throws an exception if the second argument is 0."
       , twoArg "/" $ \case
-          (Number x, Number y) | y /= 0 -> pure $ Number (x / y)
-          (Number _, Number _) -> throwErrorHere $ OtherError "Can't divide by 0"
+          (Number _, Number 0) -> throwErrorHere $ OtherError "Can't divide by 0"
```
geigerzaehler: I’m not sure this works in all cases. For example assume the rational type is implemented as an integer tuple and the literal `0` is converted to `(0, 1)`. It might be the case that we have a number that look like `(0, 2)`. This number is equal to `(0, 1)` but does not match `(0, 1)` because pattern matching works on representations.

Alltough the code might actually work it depends on the implementation of `Rational` so I’d rather err on the side of caution.

PR #362, opened 2018-12-20
Introduce 'prelude/seq'

geigerzaehler:
We move most the code from `prelude/list` that applies to both list and vectors into `prelude/seq`.

We also remove `concat` from the prelude.  We can use the `<>` primitive instead


rad/prelude/list.rad
```diff
@@ -21,22 +21,6 @@
   [ (empty? []) ==> #t ]
 )
```
MeBrei: `empty?` can now be deleted from `prelude/list`

PR #363, opened 2018-12-21
Add 'length', 'seq?', and 'starts-with?'

geigerzaehler:
* Add `prelude/seq/starts-with?`
* Add `prelude/seq/seq?`
* Add primitive function `length` that works for lists, vectors, and strings.


jameshaydon:
> Returns `#t` if `x` is a list or a vector.

It seems like we now have several seq functions (but not all) working for strings (e.g. `take` but not `drop`, now `length`, ...). Should we make strings a seq? They are in clojure.

geigerzaehler:
I think we should make strings sequences. But until all sequence primitives acutally support strings `(seq? "")` should return `#f`.

jameshaydon:
Yes let's leave it for another PR (I have created an issue).

PR #364, opened 2018-12-21
Add '/prefix' matcher

geigerzaehler:
On top of #363.

Add `/prefix` matcher for sequences and strings.


rad/prelude/seq.rad
```diff
@@ -100,3 +100,47 @@
   [ (starts-with? [] []) ==> #t ]
   [ (starts-with? [] [1]) ==> #f ]
   )
+
+(def /prefix
+  "Matches sequences that start with `prefix` and bind the rest of that
```
jameshaydon: I would say "and uses `rest-pat` to match the rest of the sequence".

PR #365, opened 2018-12-27
adds additional instructions from fresh ubuntu install

hxrts:
- req `libtinfo-dev`
- memory warning for `stack install`

### DIFF
--- a/README.md
+++ b/README.md
@@ -13,7 +13,7 @@ contain a lot more information on `radicle`.
 
 ## Installation
 
-You will need [`stack`](https://docs.haskellstack.org/en/stable/install_and_upgrade/) `libtinfo-dev`, and `libpq-dev` installed.
+You will need [`stack`](https://docs.haskellstack.org/en/stable/install_and_upgrade/) and `libpq-dev` installed.
 
 Then run:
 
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -13,12 +13,16 @@ contain a lot more information on `radicle`.
 
 ## Installation
 
-You need [stack](https://docs.haskellstack.org/en/stable/install_and_upgrade/) and `libpq-dev`.
+You will need [`stack`](https://docs.haskellstack.org/en/stable/install_and_upgrade/) `libtinfo-dev`, and `libpq-dev` installed.
```
jameshaydon: This is quite Ubuntu-specific, and also duplicates some of the info in:

http://docs.radicle.xyz/en/latest/guide/GettingStarted.html#building-from-source

I think we should make the info at that location better, and link to there from the readme.

hxrts: sounds reasonable. should I close this PR then?

jameshaydon: No, just move the edits to this file instead: `docs/source/guide/GettingStarted.rst`.

hxrts: cool, just moved.

PR #366, opened 2018-12-28
Fzf preview

jkarni:
Adds the function `fzf-select-with-preview!`, which allows keys to be previewed in a way defined by a preview function.

Uses that to give previews to issues:

preview

Navigating issues with `fzf` remains quite responsive (at least when `radicle` is compiled with optimizations).


hxrts:
yesss

PR #367, opened 2018-12-28
Exports ls! and modify-file!

jkarni:
Fixes radicle-issues 130.


hxrts:
perfect. appreciate you fixing that so quickly. was kind of expecting this would wait until after break. now I get to play with it a little bit over vacation though so ty :)

also good to know they just weren't declared as exports, will keep in mind.

PR #372, opened 2019-01-08
Issue listing filters + closing issues

jameshaydon:
Adds some filters for listing issues (`--open --closed --new`), and adds a command for closing issues.


bin/rad-issues
```diff
@@ -222,14 +254,33 @@
...
+      (match v
+             (/cons "list" (/cons 'c-n 'os)) [:just {ch-name c-n
+                                                     opts    (list-options os)}]
+             _ :nothing))))
+
 (def args (get-args!))
 (match args
+  (/list-cmd 'chain-name 'options) (list-items chain-name options)
   (/cmd-1 "list" 'chain-name) (list-items chain-name :nothing)
   (/cmd-1 "list-new" 'chain-name) (list-items chain-name (read-file-key! ".radicle-issues" :last-mark-done))
```
MeBrei: list and list-new can now be removed?

jameshaydon: Indeed

MeBrei:
I tested it locally and `bin/rad-issues list` with no chain-name & no options throws `Unknown command: list` and with options, but no chain-name throws `Unknown identifier: list-issues` (the second one might be a pre-existing problem though as it takes the option as the name of the chain)

PR #373, opened 2019-01-09
[NOMERGE] Strings are seqs and other seq improvements

jameshaydon:
- Strings work for all `seq` functions.
- `seq` functions work for more inputs, e.g. `zip` now works when one input is a list and the other is a vector.


rad/prelude/seq.rad
```diff
@@ -1,54 +1,79 @@
...
+  `xs`."
+  (fn [f xs]
+    (foldr (fn [x acc] (cons (f x) acc))
+           (empty-seq xs)
+           xs)))
+
+(:test "map"
+       [ (map (fn [x] (+ x 1)) (list 1 2 3)) ==> (2 3 4)]
+       [ (map (fn [x] (+ x 1)) [1 2 3]) ==> [2 3 4]]
+       [ (map (fn [x] (if (eq? x "f") "b" x)) "foo") ==> ["b" "o" "o"]]
```
geigerzaehler: This feels wrong to me. Should we make `map` work only for vectors and lists?

src/Radicle/Internal/PrimFns.hs
```diff
@@ -483,11 +480,12 @@ purePrimFns = fromList $ allDocs $
       )
     , ( "member?"
       , "Given `v` and structure `s`, checks if `x` exists in `s`. The structure `s`\
-        \ may be a list, vector or dict. If it is a list or a vector, it checks if `v`\
+        \ may be a sequence or dict. If it is a sequence, it checks if `v`\
```
geigerzaehler: Wouldn’t it be clearer to have to have a `elem?` prinfn and a `member?` function that only works for dicts?

jameshaydon:
Unfortunately removing `map` from the primfns breaks the issues chain :roll_eyes:

PR #374, opened 2019-01-09
RFC: radicle-daemon

jameshaydon:
[render](https://github.com/oscoin/radicle/blob/rfc/daemon/rfcs/0003-radicle-daemon.rst)

### DIFF
--- a/rfcs/0003-radicle-daemon.rst
+++ b/rfcs/0003-radicle-daemon.rst
@@ -43,10 +43,10 @@ requests on a specific port. The API uses JSON.
      "last_input_id": 55}
 
   The daemon works out if it must query a remote or local machine based on the
-  existence of a protocol (e.g. ``http://`` or ``ipfs://``). If there is no
-  protocol then the query refers to a locally maintained machine. Otherwise it
-  will fetch inputs from a remote machine using the correct protocol (but only
-  ``http`` is supported initially).
+  protocol (``radicle-local://`` versus ``http://`` or ``ipfs://``). If there is
+  no protocol then the query refers to a locally maintained machine. Otherwise
+  it will fetch inputs from a remote machine using the correct protocol (but
+  only ``http`` is supported initially).
 
   The fields ``keep_input``, ``keep_polling``, ``poll_period`` are optional.
   ``keep_input`` controls how long (seconds) the daemon will cache the inputs
@@ -83,7 +83,9 @@ Alternatives
 Unresolved question
 --------------------
 
-N/A
+- The daemon might be used by multiple apps to access data about the same remote
+  machine. In this case something should prevent one app disabling the
+  cache/polling settings of the other.
 
 Implementation
 ---------------
--- a/rfcs/0003-radicle-daemon.rst
+++ b/rfcs/0003-radicle-daemon.rst
@@ -22,10 +22,6 @@ Motivation
   inputs to the owner for inclusion into the machine. This daemon
   would listen for inputs on an IPFS pubsub channel.
 
-- The daemon also allows apps not written in radicle or Haskell to
-  have access to a radicle interpreter by maintaining *local
-  machines*.
-
 Proposal
 ---------
 
@@ -37,7 +33,8 @@ listens for HTTP requests on a specific port. It exposes a JSON API.
 **Endpoints:**
 
 Using any ``/machines`` endpoint will trigger the daemon subscribing
-and polling for activity on the machine, as described pubsub_.
+and polling for activity on the machine, as described pubsub_. Any new
+inputs received will be pinned.
 
 - *Query* a machine, that is, run an expression against it.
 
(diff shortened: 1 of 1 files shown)


cloudhead:
> Just use state serialisation, but this currently looks like a complex refactor of the radicle codebase.

Sounds like this won't use state serialisation then? If it doesn't, then it means that restarting the daemon will be very expensive? Probably not a problem right now, but eventually we'll need to be able to recover from a non-empty state when the chains start to become very long, so just something to think about.

rfcs/0003-radicle-daemon.rst
```diff
@@ -0,0 +1,98 @@
...
+- Query a machine, that is, run an expression against it.
+
+  ``POST /:machine_url/query``
+
+  .. code-block:: json
+
+    {"expression": "(get-state)",
+     "keep_input": 259200,
+     "keep_polling": 3600,
+     "poll_period": 30}
```
cloudhead: Should all keys other than "expression" be optional? Otherwise I guess new queries would override old ones even if the new query doesn't care? Eg.

```
App 1:
POST /foo/query
{"expression": "(get-state)", "keep_polling": 3600}

App 2:
POST /foo/query
{"expression": "(get-state)", "keep_polling": 0} // Ok, no longer polling!

App 1:
POST /foo/query
{"expression": "(get-state)", "keep_polling": 3600} // Woops, this will be slow because the state is stale!

Comment (cloudhead) [reply rfcs/0003-radicle-daemon.rst]: Another way around this would be for apps or queries to have a unique token:
```
POST /foo/query?token=app1
{"expression": "(get-state)", "keep_polling": 3600}

POST /foo/query?token=app2
{"expression": "(get-state)", "keep_polling": 0}

POST /foo/query?token=app1
{"expression": "(get-state)", "keep_polling": 3600} // Fresh state because we didn't stop polling for app1!
```

Comment (jameshaydon) [reply rfcs/0003-radicle-daemon.rst]: I like the token option.

Comment (geigerzaehler) [reply rfcs/0003-radicle-daemon.rst]: I also like the token option. To keep the RFC small we should add it to the “Open Questions” section.

File: rfcs/0003-radicle-daemon.rst [hunk 2]
```diff
@@ -0,0 +1,98 @@
...
+  existence of a protocol (e.g. ``http://`` or ``ipfs://``). If there is no
+  protocol then the query refers to a locally maintained machine. Otherwise it
+  will fetch inputs from a remote machine using the correct protocol (but only
+  ``http`` is supported initially).
+
+  The fields ``keep_input``, ``keep_polling``, ``poll_period`` are optional.
+  ``keep_input`` controls how long (seconds) the daemon will cache the inputs
+  for. ``keep_polling`` and ``poll_period`` control how long (seconds) it should
+  keep polling for new inputs for (after the last interaction) and the amount of
+  time between each poll.
```
jkarni: One thing we discussed offline is that we can have the polling/caching in `rad-server` as well (i.e., not just the daemon). This would mean servers act a lot like pubs in ssb, caching things on request. That in turn means the authoritative server could go offline, but clients could still get read-only access to the data. Moreover, if you want to replicate your chain, you could just ask another server to poll it.

But this introduces an attack vector, since people could ask server to poll a lot of things.

jameshaydon:
@cloudhead Agreed, serialisation should still be on the roadmap.

rfcs/0003-radicle-daemon.rst
```diff
@@ -0,0 +1,98 @@
...
+
+- Get the list of inputs of a remote machine:
+
+  ``GET /:machined_url/inputs/:id``
```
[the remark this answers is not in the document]
jameshaydon: I agree it isn't needed for diff-app, so I'll remove it.

rfcs/0003-radicle-daemon.rst
```diff
@@ -0,0 +1,144 @@
...
+- For all machines for which the daemon is a *reader*, it will poll
+  IPNS for new inputs at a low frequency. Furthermore, it will
+  subscribe to the machine's IPFS pubsub topic and listen for messages
+  with type ``"new_input"``. When such a message is received, it will
+  start polling for new inputs at a high frequency. If none are
+  detected after a certain time period the daemon will assume it was a
+  late pubsub message and return to low-frequency polling.
+
+  When new inputs are detected, the materialised machine state is
+  updated, ready for query requests.
```
jkarni: Something which we discussed elsewhere is that if the daemon is publicly-available, this same endpoint can be queried on it in order to get the data replicated outside of NAT/firewall, and for there to be a relay in the pubsub channel.

rfcs/0003-radicle-daemon.rst
```diff
@@ -0,0 +1,144 @@
...
+and polling for activity on the machine, as described pubsub_. Any new
+inputs received will be pinned.
+
+- *Query* a machine, that is, run an expression against it.
+
+  ``POST /machines/:machine_id/query "(get-state)"``
+
+  Results in something like: ``"{:apples 22 :oranges 42}"``.
+
+  The daemon will check for new inputs and then run the expression in
```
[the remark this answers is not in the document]
jameshaydon: The pubsub channel is not reliable, and the polling may be infrequent.

rfcs/0003-radicle-daemon.rst
```diff
@@ -0,0 +1,144 @@
...
+     "expression": "(add-number 42)"}
+  
+  to the IPFS pubsub topic associated to the machine and wait for a
+  message with the same nonce:
+
+  .. code-block:: json
+
+    {"type": "new_input",
+     "nonce": "abc123",
+     "expression": "(add-number 42)"}
```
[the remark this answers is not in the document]
jameshaydon: Will remove.

PR #376, opened 2019-01-10
Handle recursive functions at call site

geigerzaehler:
Instead of injecting the recursive reference to a function into the closure when defining the recursive function we inject it into the environment when calling the function. This should eliminate circular references from `Env`.

### DIFF
--- a/src/Radicle/Internal/Core.hs
+++ b/src/Radicle/Internal/Core.hs
@@ -252,6 +252,12 @@ data ValueF r =
     -- The value of an application of a lambda is always the last value in the
     -- body. The only reason to have multiple values is for effects.
     | LambdaF [Ident] (NonEmpty r) (Env r)
+    -- | Like 'LambdaF' but indicates a function that can call itself
+    -- recursively.
+    --
+    -- The first argument is the name for recursive calls to the
+    -- function in the body.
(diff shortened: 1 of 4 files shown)


src/Radicle/Internal/Core.hs
```diff
@@ -722,8 +735,7 @@ specialForms = Map.fromList $ first Ident <$>
       val' <- baseEval val
       case val' of
         Lambda is b e -> do
```
jameshaydon: Here there should also be a case for `LambdaRec` too?

geigerzaehler: I’m not sure. I think the only way you can have a `LambdaRec` here is with the following code
~~~clojure
(def-rec foo (fn [] nil))
(def-rec bar foo)
~~~
But this does not really make

jameshaydon: I think it could happen quite reasonably:

```
(def-rec mock-foo ...)
(def-rec real-foo ...)

(def-rec foo (if (read-ref debug-mode?) mock-foo real-foo))
```

So I think it's fine to say that the user should use `def` rather than `def-rec` in this case, but let's then give an informative error.

geigerzaehler: Good point. Can you have another look

PR #381, opened 2019-01-11
Add IPFS machine backend

geigerzaehler:
* Add a `MachineBackend` implementation based on IPFS
* Extend `test/machine-backends.rad` to test IPFS backend against real IPFS daemon
* Add commands `rad-ipfs`, `rad-ipfs-daemon`, and `rad-ipfs-machine-create`

Follow-up

* Replace `Wreq` with `servant-client` and `ipfs-api` from   https://github.com/oscoin/ipfs/tree/master/ipfs-api
* Move aome code to `ipfs-api`


package.yaml
```diff
@@ -30,12 +30,16 @@ library:
   - generics-eot
   - haskeline
   - http-client
+  - ipld-cid
   - megaparsec
   - mtl
+  - lens
```
cloudhead: Are we using Prisms? In the spirit of smaller build times, might be worth using micro-lens here, if we aren't using Prisms etc.

cloudhead: Though I just see you're using `wreq`, which I think depends on `lens` anyway, so maybe ignore..

src/Radicle/Internal/MachineBackend/Ipfs.hs
```diff
@@ -0,0 +1,318 @@
...
+    toJSON MachineEntry{..} =
+        let code = T.intercalate "\n" $ map renderCompactPretty entryExpressions
+        in Aeson.object
+            [ "expression" .= code
+            , "previous" .= ipldLink entryPrevious
+            ]
+
+-- | Given a CID @"abc...def"@ it returns a IPLD link JSON object
+-- @{"/": "abc...def"}@.
+ipldLink :: CID -> Aeson.Value
```
kim: This would be nice to have in [oscoin/ipfs](https://github.com/oscoin/ipfs) somewhere

geigerzaehler: Yes indeed. There is a couple of things we should backport, like the response types.

package.yaml
geigerzaehler: I’m switching to `microlens` in our code. Unfortunately `wreq` still has the `lens` dependency. We will get of `wreq` soon.

snapshot.yaml
```diff
@@ -9,6 +9,19 @@ name: radicle-deps-13.2
...
+- serialise-0.2.1.0
+- git: https://github.com/oscoin/ipfs.git
+  commit: 19b25fa4d0003b5c2027c8cc1b8a3c36ac2f9f60
+  subdirs:
+  - ipld-cid
+  - binary-varint
+  - multibase
+  - multihash-cryptonite
+
+flags:
```
kim: Dang. bump oscoin/ipfs#7

src/Radicle/Internal/MachineBackend/Ipfs.hs
```diff
@@ -0,0 +1,318 @@
...
+    -- | JSON response from the IPFS Api cannot be parsed. First
+    -- argument is the request path, second argument the JSON parsing
+    -- error
+    | IpfsExceptionInvalidResponse Text Text
+    -- | The IPFS daemon is not running.
+    | IpfsExceptionNoDaemon
+    deriving (Show, Eq)
+
+instance Exception IpfsException where
+    displayException (IpfsException msg) = "ipfs: " <> toS msg
```
kim: TIL `displayException` is a thing

src/Radicle/Internal/MachineBackend/Ipfs.hs
```diff
@@ -0,0 +1,318 @@
+-- | This module implements an IPFS machine backend.
+--
+-- Radicle Machines are identified by IPNS IDs. The IPNS ID of a
+-- machine points to a linked listed of expressions encoded as content
```
jameshaydon:
```suggestion
-- machine points to a linked list of expressions encoded as content
```

PR #386, opened 2019-01-16
Add docker-compose file to run all tests

geigerzaehler:
Add `test/docker-compose.yaml` that provides all services required to
run the machine backend tests.

Also includes a commit that adds the `RAD_` prefix to an environment
variable.

### DIFF
--- a/cloudbuild.yaml
+++ b/cloudbuild.yaml
@@ -156,7 +156,7 @@ steps:
     - |
       set -euxo pipefail
       stack exec -- radicle - <<<'(load! "rad/examples/counter.rad") (counter/run-test)'
-      IPFS_API_URL=http://ipfs-test-network:5001 stack exec -- radicle test/machine-backends.rad radicle-server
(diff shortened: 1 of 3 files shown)


cloudbuild.yaml
```diff
@@ -156,7 +154,7 @@ steps:
     - |
       set -euxo pipefail
       stack exec -- radicle - <<<'(load! "rad/examples/counter.rad") (counter/run-test)'
-      IPFS_API_URL=http://ipfs-test-network:5001 stack exec -- radicle test/machine-backends.rad radicle-server
+      RAD_IPFS_API_URL=http://ipfs-test-network:5001 stack exec -- radicle test/machine-backends.rad radicle-server
```
kim: Is `IPFS_API_URL` somehow taken, eg. by the ipfs CLI? I made the remote helper respect that and thought it might come in handy to have it set once when running the whole stack.

geigerzaehler: I think it makes sense to namespace it because the Radicle network is distinct from the main IPFS network. I would like to avoid the situation where some other tool that talks to the main network understands the IPFS_API_URL variable and gets confused.

PR #394, opened 2019-01-21
Add note about NAT/firewalled computers.

jkarni:


rfcs/0003-radicle-daemon.rst
```diff
@@ -123,7 +123,23 @@ inputs received will be pinned.
   updated IPNS record. If this is successful, the daemon posts a
   message of type ``"new_input"`` with the same nonce to the
   machines's pubsub topic.
-  
+
+**Inaccessible computers and relayers**
+
+Many computers will be behind firewalls, or NATs that IPFS/pubsub cannot
+traverse. People who host writer chains on such computers can use
```
geigerzaehler: The term “host writer chains” seems a bit opaque to me. I think it would be more descriptive to talk about the IPFS daemon that controls a machine’s IPNS ID

rfcs/0003-radicle-daemon.rst
```diff
@@ -123,7 +123,23 @@ inputs received will be pinned.
...
+publicly-accessible daemons ("relayers") to mitigate this problem. By querying
+their own RSMs via such daemons, they will:
+
+- Initiate pinning by a publicly-available computer; and
+- Make the publicly-available computer join their pubsub channel, thus acting
+  as a relay.
```
geigerzaehler: I assume this is a temporary solution. Can we add a paragraph that mentions that?

PR #397, opened 2019-01-22
Rad path

jkarni:
For the alpha, we need radicle to work when run from anywhere, not just the radicle repo. (This is particularly important for testing the project app, since we need to cd elsewhere etc.) 

This work to introduce a rad search path (and RADPATH). 

The proposed semantics:

- First search `data-files` dir
- Then search RADPATH
- Then search current dir

This is like Python, with the difference that Python searches the current dir before PYTHONPATH. My reasoning is that changing the order makes it harder to create an attack by adding certain files to a repo.


geigerzaehler:
How about we only use `data-dir` as the default for `RADPATH`?

I would avoid using the cwd to find modules. With that your commands might stop working if called from a different directory. To use local module files we should resolve them relative to the requiring source file like Node does.

jkarni:
> How about we only use data-dir as the default for RADPATH?

How do you mean? As in, if RADPATH is not set, then use data-dir and nothing else? 

> To use local module files we should resolve them relative to the requiring source file like Node does.

That's a good idea, but I think a bigger implementation. Okay if I do it with cwd now (plus data-dir/radpath) and create a new issue?

geigerzaehler:
> How do you mean? As in, if RADPATH is not set, then use data-dir and nothing else?

If `RADPATH` is not set use `getDataDir` from `Paths_radicle` which is determined by the distribution and package. If `RADPATH` is set, use the value as is.

> That's a good idea, but I think a bigger implementation. Okay if I do it with cwd now (plus data-dir/radpath) and create a new issue?

That makes sense. It would be nice to avoid the brittle cwd system so we don’t shoot ourselves in the foot. You can always work around it by changing `RADPATH`.

bin/rad-issues
```diff
@@ -1,8 +1,8 @@
 #!/usr/bin/env radicle
 
-(load! "rad/prelude.rad")
-(load! "rad/monadic/issues.rad")
-(load! "rad/prelude/io-utils.rad")
+(load! (find-module-file! "rad/prelude.rad"))
```
geigerzaehler: The `rad` prefix feels kind of redundant. Can’t we do away with it?

src/Radicle/Internal/Effects/Capabilities.hs
```diff
@@ -137,8 +140,46 @@ instance ReadFile IO where
...
+            Just rp -> pure $ splitPath rp
+    let searchPath = radPathDirs ++ ["."]
+    let doFind xs = case xs of
+            [] -> pure Nothing
+            x:xs' -> do
+                exists <- doesFileExist (x </> file)
+                if exists then pure (Just $ x </> file) else doFind xs'
+    doFind searchPath
+  where
+    splitPath p = reverse $ splitPathH p []
```
geigerzaehler: We could use [`splitSearchPath`](https://hackage.haskell.org/package/filepath-1.4.2.1/docs/System-FilePath-Posix.html#v:splitSearchPath)

src/Radicle/Internal/Effects/Capabilities.hs
```diff
@@ -137,8 +140,46 @@ instance ReadFile IO where
...
+
 putStrLnS :: (Stdout m) => Text -> m ()
 putStrLnS t = putStrS t >> putStrS "\n"
 
 modifyEnvS :: (GetEnv m r, SetEnv m r) => (Env r -> Env r) -> m ()
 modifyEnvS f = getEnvS >>= setEnvS . f
+
+-- | Importing works as follows:
+--
+--    (1) If RADPATH exists, search there
```
geigerzaehler: Could you please add that `RADPATH` has the same format as PATH with multiple paths separated by a colon.

src/Radicle/Internal/Effects/Capabilities.hs
```diff
@@ -137,8 +140,46 @@ instance ReadFile IO where
 instance {-# OVERLAPPABLE #-} ReadFile m => ReadFile (Lang m) where
     readFileS = lift . readFileS
 
+class (Monad m) => FileModule m where
+    fileModuleS :: Text -> m (Either Text Text)
```
geigerzaehler: I would prefer `m (Maybe Text)` as the return type and let the caller handle missing files.

src/Radicle/Internal/Effects/Capabilities.hs
```diff
@@ -137,8 +140,46 @@ instance ReadFile IO where
...
 putStrLnS t = putStrS t >> putStrS "\n"
 
 modifyEnvS :: (GetEnv m r, SetEnv m r) => (Env r -> Env r) -> m ()
 modifyEnvS f = getEnvS >>= setEnvS . f
+
+-- | Importing works as follows:
+--
+--    (1) If RADPATH exists, search there
+--    (2) Otherwise search in data-dir
+--    (3) Finally we search in the current directory.
```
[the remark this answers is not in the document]
jkarni: It is though

src/Radicle/Internal/Effects/Capabilities.hs
jkarni: Nice

src/Radicle/Internal/Effects/Capabilities.hs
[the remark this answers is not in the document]
jameshaydon: Can we keep cwd *until* we have searching relative to the importing file?

bin/rad-issues
jkarni: Done

src/Radicle/Internal/Effects/Capabilities.hs
jkarni: Done.

src/Radicle/Internal/Effects/Capabilities.hs
jkarni: Done

src/Radicle/Internal/Effects/Capabilities.hs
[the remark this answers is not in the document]
jkarni: > Can we keep cwd until we have searching relative to the importing file?

That's what I'd prefer too.

src/Radicle/Internal/Effects.hs
```diff
@@ -322,12 +323,29 @@ replPrimFns sysArgs = fromList $ allDocs $
...
+    , ( "find-module-file!"
+      , "Find a file according to radicle search path rules. These are: \
+        \ 1) If RADPATH is set, first search there; \
+        \ 2) If RADPATH is not set, search in the distribution directory \
+        \ 3) If the file is still not found, search in the current directory."
+      , oneArg "find-module-file!" $ \case
+          String filename -> do
+            t_ <- fileModuleS filename
+            case t_ of
+                Nothing -> throwError . toLangError . OtherError $ "File not found: " <> filename
```
geigerzaehler: I got confused by this when I tested it. Can we make this “File not found in RADPATH: filename”? Same for `file-module!`.

package.yaml
```diff
@@ -20,8 +20,7 @@ ghc-options:
   - -Wredundant-constraints
   - -threaded
 
-data-files:
-- rad/**/*.rad
+data-dir: rad
```
geigerzaehler: This does not seem to work for me. `stack build` will not copy the files into the right location.

src/Radicle/Internal/Effects/Capabilities.hs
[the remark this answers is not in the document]
geigerzaehler: Okay, let’s also use the cwd.

src/Radicle/Internal/Effects/Capabilities.hs
```diff
@@ -137,8 +140,42 @@ instance ReadFile IO where
...
+--
+-- RADPATH should be a colon-separated list of directories (just like PATH).
+findModule :: FilePath -> IO (Maybe FilePath)
+findModule file = do
+    mrp <- lookupEnv "RADPATH"
+    radPathDirs <- case mrp of
+            Nothing -> (\x -> [x </> "rad"]) <$> getDataDir
+            Just rp -> pure $ splitSearchPath rp
+    let searchPath = radPathDirs ++ ["."]
```
geigerzaehler: I don’t think this works right now. Should be `./rad`

jkarni: Instead of "."? Not sure I get it.

package.yaml
jkarni: Explains why CI was failing!

src/Radicle/Internal/Effects.hs
jkarni: Done.

jkarni:
> Could we also make sure that the CI tests use a different working directory to ensure that the RADPATH actually works?

Since it's run from `workspace` but the files are in `workspace/rad`, the tests already use the RADPATH semantics. But I've added explicit RADPATHs (which override data-files) to test that.

jkarni:
@geigerzaehler I think I've addressed all comments (or asked further questions). Let me know if this is good to merge.

PR #398, opened 2019-01-22
Implement basic rad top-level command

cloudhead:
Implements a top-level entry point command, `rad`.

`rad`, `rad help`, `rad --help` and `rad -h` all just print the list of available commands.

Notes:
* I started off with optparse-applicative, but it was getting too complicated, probably in part due to my lack of knowledge of the library, so I opted for a quick and dirty approach. It'll be easy to refactor to optparse-applicative when/if we need more complex parsing.
* Presumably doesn't work on some platforms due to the use of `executeFile` (exec)
* Checks the `bin` dir for now for sub-commands, though eventually we might want to use `/usr/local/libexec`.
* ~~The `rad version` command is a placeholder, we'll have to have it pick up the actual version somehow, maybe as another changeset.~~

### DIFF
--- a/exe/Rad.hs
+++ b/exe/Rad.hs
@@ -9,10 +9,11 @@ import           Protolude
 import           Data.List (isPrefixOf)
 import           Data.String (String)
 import qualified Data.Text as T
+import           Data.Version (showVersion)
 import           System.Directory
 import qualified System.Posix.Process as Posix
 
-import           Paths_radicle (getBinDir)
+import           Paths_radicle (getBinDir, version)
 
 -- | All command filenames are prefixed with this.
 radPrefix :: String
@@ -26,7 +27,7 @@ runCommand [] = do
     putStrLn @Text "Available rad commands:"
     putStr $ T.unlines $ ["   " <> T.pack c | c <- cmds]
 runCommand ("version":_) =
-    putStrLn ("rad version 1.0" :: Text)
+    putStrLn ("rad version " <> toS (showVersion version) :: Text)
 runCommand ("help":_) =
     runCommand []
 runCommand ("--help":_) =
--- a/exe/Rad.hs
+++ b/exe/Rad.hs
@@ -1,0 +1,64 @@
+{-# LANGUAGE TypeApplications #-}
+-- | Top-level radicle CLI. Calls sub-commands.
+module Rad
+    ( main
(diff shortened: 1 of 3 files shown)


exe/Rad.hs
```diff
@@ -0,0 +1,64 @@
...
+radPrefix :: String
+radPrefix = "rad-"
+
+runCommand :: [String] -> IO ()
+runCommand [] = do
+    bin <- getBinDir
+    cmds <- availableCommands [bin]
+    putStr @Text "usage: rad <command> [<args>]\n\n"
+    if null cmds then do
+        putStrLn @Text "No available rad commands."
```
geigerzaehler: Maybe it would make sense to list the `version` and `help` commands here?

package.yaml
```diff
@@ -87,6 +87,19 @@ tests:
...
+    main: Rad.hs
+    source-dirs: exe
+    other-modules: [Paths_radicle]
+    dependencies:
+    - directory
+    - unix
+    - text
+    when:
+      - condition: impl(ghcjs)
+        buildable: false
```
geigerzaehler: You can remove the `when` clause. We removed the GHCJS build

exe/Rad.hs
```diff
@@ -0,0 +1,64 @@
...
+    else do
+        putStrLn @Text "Available rad commands:"
+        putStr $ T.unlines $ ["   " <> T.pack c | c <- cmds]
+runCommand ("version":_) =
+    putStrLn ("rad version 1.0" :: Text)
```
geigerzaehler: You can use `version` from `Paths_radicle`. See the [cabal docs](https://cabal.readthedocs.io/en/latest/developing-packages.html?highlight=caret#accessing-the-package-version).

cloudhead: Ah perfect!

exe/Rad.hs
cloudhead: Yeah, for some reason `git help -a` doesn't list `version` but it does list `help`.

exe/Rad.hs
cloudhead: 9ef9734

exe/Rad.hs
cloudhead: 85fdcc3

PR #399, opened 2019-01-22
Radicle daemon

jameshaydon:
As per [RFC](https://github.com/oscoin/radicle/blob/master/rfcs/0003-radicle-daemon.rst).

- [x] Polling.
- [x] Thread safe machine cache.
- [ ] Filelock to avoid several daemons running on one machine.

### DIFF
--- a/exe/Daemon.hs
+++ b/exe/Daemon.hs
@@ -1,0 +1,583 @@
+-- | The radicle-daemon; a long-running background process which
+-- materialises the state of remote IPFS machines on the user's PC, and
+-- writes to those IPFS machines the user is an owner of.
+--
+-- See
+-- <https://github.com/oscoin/radicle/blob/master/rfcs/0003-radicle-daemon.rst
+-- the RFC>.
+module Daemon where
+
+import           Protolude hiding (fromStrict, option, poll)
+
+import qualified Data.Aeson as A
+import           Data.ByteString.Lazy (fromStrict)
+import qualified Data.Map.Strict as Map
+import qualified Data.Sequence as Seq
+import qualified Data.Text as T
+import qualified Data.Time.Clock.System as Time
+import           Network.Wai.Handler.Warp (run)
+import           Options.Applicative
+import           Servant
+import           System.Directory (doesFileExist)
+
+import           Radicle.Daemon.Common hiding (logInfo)
+import qualified Radicle.Daemon.Common as Common
+import qualified Radicle.Internal.UUID as UUID
+
+import           Radicle hiding (Env)
+import           Radicle.Daemon.HttpApi
+import           Radicle.Daemon.Ipfs
+import qualified Radicle.Internal.CLI as Local
+import qualified Radicle.Internal.ConcurrentMap as CMap
+
+
+-- TODO(james): Check that the IPFS functions are doing all the
+-- necessary pinning.
+
+-- * Types
+
+data MachineError
+  = InvalidInput (LangError Value)
+  | IpfsError IpfsError
+  | AckTimeout
+  | DaemonError Text
+  | MachineAlreadyCached
+  | MachineNotCached
+
+data Error
+  = MachineError MachineId MachineError
+  | CouldNotCreateMachine Text
+
+displayError :: Error -> (Text, [(Text,Text)])
+displayError = \case
+  CouldNotCreateMachine e -> ("Could not create IPFS machine", [("error", e)])
+  MachineError id e -> let mid = ("machine-id", getMachineId id) in
+    case e of
+      InvalidInput err -> ("Invalid radicle input", [mid, ("error", renderCompactPretty err)])
+      DaemonError err -> ("Internal error", [mid, ("error", err)])
+      IpfsError e' -> case e' of
+        IpfsDaemonError err -> ("There was an error using IPFS", [mid, ("error", toS (displayException err))])
+        InternalError err -> ("Internal error", [mid, ("error", err)])
+        NetworkError err -> ("There was an error communicating with the IPFS daemon", [mid, ("error", err)])
+      AckTimeout -> ("The writer appears to be offline", [mid])
+      MachineAlreadyCached -> ("Tried to add already cached machine", [mid])
+      MachineNotCached -> ("Machine was not found in cache", [mid])
+
+logDaemonError :: MonadIO m => Error -> m ()
+logDaemonError (displayError -> (m, xs)) = logErr m xs
+
+type Follows = Map MachineId ReaderOrWriter
+
+-- * Main
+
+data Opts = Opts
+  { port       :: Int
+  -- TODO(james): temporary, just for testing.
+  , filePrefix :: Text
+  , debug      :: Bool
+  }
+
+opts :: Parser Opts
+opts = Opts
+    <$> option auto
+        ( long "port"
+       <> help "daemon port"
+       <> metavar "PORT"
+       <> showDefault
+       <> value 8909
+        )
+    <*> strOption
+        ( long "filePrefix"
+       <> help "file prefix"
+       <> metavar "PREFIX"
+       <> showDefault
+       <> value ""
+        )
+    <*> switch
+        ( long "debug"
+       <> help "enable debug logging"
+       <> showDefault
+        )
+
+type FollowFileLock = MVar ()
+
+data LogLevel = Normal | Debug
+  deriving (Eq, Ord)
+
+data Env = Env
+  { followFileLock :: FollowFileLock
+  , followFile     :: FilePath
+  , machines       :: IpfsMachines
+  , logLevel       :: LogLevel
+  }
+
+newtype Daemon a = Daemon { fromDaemon :: ExceptT Error (ReaderT Env IO) a }
+  deriving (Functor, Applicative, Monad, MonadError Error, MonadIO, MonadReader Env)
+
+runDaemon :: Env -> Daemon a -> IO (Either Error a)
+runDaemon env (Daemon x) = runReaderT (runExceptT x) env
+
+main :: IO ()
+main = do
+    Opts{..} <- execParser allOpts
+    followFileLock <- newMVar ()
+    followFile <- Local.getRadicleFile (toS filePrefix <> "daemon-follows")
+    machines <- Chains <$> CMap.empty
+    let env = Env{ logLevel = if debug then Debug else Normal, ..}
+    follows <- readFollowFileIO followFileLock followFile
+    initRes <- runDaemon env (init follows)
+    case initRes of
+      Left err -> logErr "Init failed" [] >> logDaemonError err
+      Right _ -> do
+        polling <- async $ initPolling env
+        let app = serve daemonApi (server env)
+        Common.logInfo "Start listening" [("port", show port)]
+        serv <- async $ run port app
+        exc <- waitEitherCatchCancel polling serv
+        case exc of
+          Left (Left err) -> logErr "Polling failed with an exception" [("error", toS (displayException err))]
+          Left (Right void') -> absurd void'
+          Right (Left err) -> logErr "Server failed with an exception" [("error", toS (displayException err))]
+          Right (Right ()) -> logErr "Server stopped (this should not happen)" []
+  where
+    allOpts = info (opts <**> helper)
+        ( fullDesc
+       <> progDesc "Run the radicle daemon"
+       <> header "radicle-daemon"
+        )
+
+type IpfsMachine = Chain MachineId MachineEntryIndex TopicSubscription
+type IpfsMachines = Chains MachineId MachineEntryIndex TopicSubscription
+
+server :: Env -> Server DaemonApi
+server env = hoistServer daemonApi nt daemonServer
+  where
+    daemonServer :: ServerT DaemonApi Daemon
+    daemonServer = (machineEndpoints :<|> newMachine) :<|> pure swagger
+    machineEndpoints id = query id :<|> send id
+
+    nt :: Daemon a -> Handler a
+    nt d = do
+      x_ <- liftIO $ runDaemon env d
+      case x_ of
+        Left err -> throwError (toServantErr err)
+        Right x  -> pure x
+
+    toServantErr err@(MachineError _ e) = case e of
+      InvalidInput e' -> err400 { errBody = fromStrict $ encodeUtf8 (renderPrettyDef e') }
+      AckTimeout -> err504 { errBody = "The writer for this IPFS machine does not appear to be online." }
+      _ -> case displayError err of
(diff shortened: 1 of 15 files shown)


src/Radicle/Daemon/Common.hs
```diff
@@ -0,0 +1,66 @@
...
+
+data ReaderOrWriter = Reader | Writer
+  deriving (Generic)
+
+data Polling = HighFreq Int64 | LowFreq
+
+instance A.ToJSON ReaderOrWriter
+instance A.FromJSON ReaderOrWriter
+
+data Chain id idx subs = Chain
```
geigerzaehler: Can we replace the term “Chain” with “Machine”

src/Radicle/Daemon/Common.hs
```diff
@@ -0,0 +1,66 @@
+module Radicle.Daemon.Common
```
geigerzaehler: Can you add documentation to the values and types defined here that don’t yet have documentation?

src/Radicle/Daemon/Common.hs
geigerzaehler: What is the reason behind parameterizing the type?

src/Radicle/Internal/ConcurrentMap.hs
```diff
@@ -0,0 +1,83 @@
+module Radicle.Internal.ConcurrentMap
```
geigerzaehler: Is there a reason why we don’t use [`StmContainers.Map`](http://hackage.haskell.org/package/stm-containers-1.1.0.2/docs/StmContainers-Map.html)?

exe/Daemon.hs
```diff
@@ -0,0 +1,485 @@
+{-# OPTIONS_GHC -fno-warn-orphans #-}
+
+-- | The radicle-daemon; a long-running background process which
+-- materialises the state of remote IPFS machines on the users PC, and
+-- writes to those IPFS machines the user is an owner of.
+--
+-- See
+-- <https://github.com/oscoin/radicle/blob/master/rfcs/0003-radicle-daemon.rst
+-- the RFC>.
+module Daemon where
```
geigerzaehler: Can we split out the actual everything but the CLI parsing into a separate module?

src/Radicle/Daemon/Ipfs.hs
```diff
@@ -0,0 +1,213 @@
...
+newtype Topic = Topic { getTopic :: Text }
+
+-- | The IPFS pubsub topic associated to a machine.
+machineTopic :: MachineId -> Topic
+machineTopic (MachineId id) = Topic ("radicle:machine:" <> id)
+
+-- | Publish a 'Message' on a machine's IPFS pubsub topic.
+publish :: MachineId -> Message -> ExceptT IpfsError IO ()
+publish id msg = safeIpfs $ do
+  liftIO $ putStrLn $ "pubsub pub " <> getMachineId id <> ": " <> toS (Aeson.encode msg)
```
geigerzaehler: Should use `log`

src/Radicle/Daemon/Ipfs.hs
```diff
@@ -0,0 +1,213 @@
...
+  Ipfs.publish (getTopic (machineTopic id)) (Aeson.encode msg)
+
+-- | Subscribe to messages on a machine's IPFS pubsub topic.
+subscribeForever :: MachineId -> (Message -> IO ()) -> IO ()
+subscribeForever id messageHandler = Ipfs.subscribe topic pubsubHandler
+  where
+    Topic topic = machineTopic id
+    pubsubHandler Ipfs.PubsubMessage{..} = do
+        liftIO $ putStrLn $ "pubsub sub " <> getMachineId id <> ": " <> toS messageData
```
geigerzaehler: Should use `log`

src/Radicle/Daemon/Ipfs.hs
```diff
@@ -0,0 +1,213 @@
...
+        case decodeStrict messageData of
+            Nothing  -> putStrLn ("Cannot parse pubsub message" :: Text)
```
geigerzaehler: Should use `log`

exe/Daemon.hs
```diff
@@ -0,0 +1,485 @@
+{-# OPTIONS_GHC -fno-warn-orphans #-}
+
+-- | The radicle-daemon; a long-running background process which
+-- materialises the state of remote IPFS machines on the users PC, and
```
jkarni:
```suggestion
-- materialises the state of remote IPFS machines on the user's PC, and
```

exe/Daemon.hs
```diff
@@ -0,0 +1,485 @@
...
+    Nothing -> do
+      m <- initAsReader id
+      requestInput (chainSubscription m)
+  where
+    requestInput sub = do
+      nonce <- liftIO $ UUID.uuid
+      let isResponse = \case
+            New NewInputs{ nonce = Just nonce' } | nonce' == nonce -> True
+            _ -> False
+      asyncMsg <- liftIO $ async $ subscribeOne sub 4000 isResponse -- Waits for 4 seconds.
```
jkarni: Maybe move this `4000` to a top-level var so it's easier to see/configure.

exe/Daemon.hs
```diff
@@ -0,0 +1,485 @@
...
+        )
+    <*> strOption
+        ( long "filePrefix"
+       <> help "file prefix"
+       <> metavar "PREFIX"
+       <> showDefault
+       <> value ""
+        )
+
+type FollowFileLock = MVar ()
```
jkarni: How bad is it if two different daemons are started? Do they notice one another and quit? If not, and clobbering the file is serious, it may be worth using `flock` (via e.g. [`filelock`](http://hackage.haskell.org/package/filelock-0.1.1.2/docs/System-FileLock.html)) since MVars only work within one process.

exe/Daemon.hs
```diff
@@ -0,0 +1,485 @@
...
+daemonHandler :: Env -> (Message -> Daemon ()) -> Message -> IO ()
+daemonHandler env h msg = do
+  x_ <- runDaemon env (h msg)
+  case x_ of
+    Left err -> logDaemonError err
+    Right _  -> pure ()
+
+
+
+-- Loads a machine fresh from IPFS.
```
jkarni:
```suggestion
-- | Loads a machine fresh from IPFS.
```

exe/Daemon.hs
```diff
@@ -0,0 +1,485 @@
...
+loadMachine :: ReaderOrWriter -> MachineId -> Daemon IpfsMachine
+loadMachine mode id = do
+  (idx, is) <- ipfs id $ machineInputsFrom id Nothing
+  sub <- ipfs id $ initSubscription id
+  m <- liftIO $ emptyMachine id mode sub
+  (m', _) <- addInputs is (pure idx) (const (pure ())) m
+  insertNewMachine m'
+  pure m'
+
+-- Add inputs to a cached machine.
```
jkarni:
```suggestion
-- | Add inputs to a cached machine.
```

src/Radicle/Daemon/Common.hs
jameshaydon: It was to make it useable in both the server and the daemon. The shared code turned out to be very small though; not sure it's worth it.

exe/Daemon.hs
```diff
@@ -0,0 +1,485 @@
...
+  -> Daemon MachineEntryIndex
+  -- ^ Determine the new index.
+  -> ([Value] -> Daemon ())
+  -- ^ Performed after the chain is cached.
+  -> IpfsMachine
+  -> Daemon (IpfsMachine, [Value])
+  -- ^ Returns the updated machine and the results.
+addInputs is getIdx after m =
+  case advanceChain m is of
+    Left err -> Daemon $ ExceptT $ pure $ Left $ MachineError (chainName m) (InvalidInput err)
```
jkarni:
```suggestion
    Left err -> throwError $ MachineError (chainName m) (InvalidInput err)
```
I think?

src/Radicle/Internal/ConcurrentMap.hs
jameshaydon: First of all it required some extra stuff in `snapshot.yaml`:

```
- primitive-extras-0.7.1@sha256:23905c57089418b1a2d324cfee3e81bbd5a344a0fa56a827867b2dce275fdb5e
- stm-containers-1.1.0.2@sha256:8976df591e61f378ef29dc0443dbe2d5d34818c898093af3ca138ea1d013b5eb
- stm-hamt-1.2.0.2@sha256:18126db7bf2d9c967a6020c677b3005dd957a4c39d69aeaea3c29c90de8f6124
```

And second of all, I wanted the writer threads to intereact with IPFS serially (per machine).

exe/Daemon.hs
```diff
@@ -0,0 +1,515 @@
...
+                 , chainEvalPairs = chainEvalPairs m Seq.>< news
+                 , chainLastUpdated = t
+                 }
+      after rs
+      pure (m', rs)
+
+ipfs :: MachineId -> ExceptT IpfsError IO a -> Daemon a
+ipfs id = Daemon . mapExceptT (lift . fmap (first (MachineError id . IpfsError)))
+
+-- Do some high-freq polling for a while.
```
jkarni: The polling gets bumped back down automatically?

exe/Daemon.hs
```diff
@@ -0,0 +1,515 @@
...
+bumpPolling :: MachineId -> Daemon ()
+bumpPolling id = do
+  modifyMachine id $ \m' -> pure (m' { chainPolling = highFreq }, () )
+  logInfo Debug "Reset to high-frequency polling" [("id", getMachineId id)]
+
+-- Insert a new machine into the cache. Errors if the machine is
```
jkarni:
```suggestion
-- | Insert a new machine into the cache. Errors if the machine is
```

exe/Daemon.hs
jameshaydon: This was just to avoid separate threads in one daemon from clobbering each other. I will use `filelock` to prevent two daemons from starting up (it didn't yet because it was nice to run two for testing).

exe/Daemon.hs
jameshaydon: Yes.

exe/Daemon.hs
```diff
@@ -0,0 +1,515 @@
...
+insertNewMachine m = do
+    msVar <- asks machines
+    inserted_ <- liftIO $ CMap.insertNew id m (getChains msVar)
+    case inserted_ of
+      Just () -> pure ()
+      Nothing -> throwError (MachineError id MachineAlreadyCached)
+  where
+    id = chainName m
+
+-- Modify a machine that is already in the cache. Errors if the
```
jkarni:
```suggestion
-- | Modify a machine that is already in the cache. Errors if the
```

exe/Daemon.hs
```diff
@@ -0,0 +1,515 @@
...
+            , chainPolling = highFreq
+            }
+
+-- | High frequency polling lasts for 10 mins.
+highFreq :: Polling
+highFreq = HighFreq (10 * 60 * 1000)
+
+-- ** Reader
+
+-- Initialise the daemon service for this machine in /reader mode/.
```
jkarni:
```suggestion
-- | Initialise the daemon service for this machine in /reader mode/.
```

exe/Daemon.hs
```diff
@@ -0,0 +1,515 @@
...
+    pure m
+  where
+    onMsg :: Message -> Daemon ()
+    onMsg = \case
+      New NewInputs{..} -> do
+        _ <- refreshAsReader id
+        bumpPolling id
+      _ -> pure ()
+
+-- Freshen up a cached machine.
```
jkarni:
```suggestion
-- | Freshen up a cached machine.
```

exe/Daemon.hs
```diff
@@ -0,0 +1,515 @@
...
+
+-- ** Writer
+
+initAsWriter :: MachineId -> Daemon IpfsMachine
+initAsWriter id = do
+  m <- loadMachine Writer id
+  actAsWriter m
+  pure m
+
+-- Subscribes the daemon to the machine's pubsub topic to listen for
```
jkarni:
```suggestion
-- | Subscribes the daemon to the machine's pubsub topic to listen for
```

src/Radicle/Daemon/Ipfs.hs
```diff
@@ -0,0 +1,213 @@
...
+-- | Message sent to request the writer add inputs to the machine.
+data ReqInputs = ReqInputs
+  { nonce       :: Text
+  , expressions :: [JsonValue]
+  }
+
+-- * Ipfs helpers
+
+safeIpfs :: forall a. IO a -> ExceptT IpfsError IO a
+safeIpfs io = ExceptT $ liftIO $
```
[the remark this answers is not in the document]
geigerzaehler: Don’t need to fix this now but we should record it somewhere

src/Radicle/Daemon/Common.hs
geigerzaehler: Let’s not share code between the server and the daemon. It is very little code and we are going to kill the server. So copy paste is fine.

exe/Daemon.hs
jameshaydon: Added to tech-debt.

jkarni:
I'm inclined to think the polling mechanism will have to be reconsidered - it seems relatively ad-hoc right now. We can use something like [retry](http://hackage.haskell.org/package/retry-0.7.7.0/docs/Control-Retry.html) eventually, although I recognize that the polling is not just based on how frequently polling got new inputs, but also how frequently the new inputs are requested from the local daemon. In theory I think all of that logic should be in the `CMap` datatype (or on a new datatype that wraps it) - the API that the datatype exposes should be such that reading and writing tend to increase polling. 

I also think the mvar situation is going to get us into performance problems sooner or later. 

If that's okay, I'll add tech debt tasks related to this?

src/Radicle/Daemon/Common.hs
```diff
@@ -0,0 +1,66 @@
...
+-- The second argument is a list of key-value pairs that will be joined with "="
+-- and appended to the message.
+--
+-- @
+--      log "INFO" "the message" [("foo", "5)]
+--      -- prints "INFO   the message foo=5"
+-- @
+log :: MonadIO m => Text -> Text -> [(Text, Text)] -> m ()
+log typ msg dat = do
+    putStrLn $ typ <> "  " <> msg <> datString
```
jkarni: Why aren't we just using an off-the-shelf logger?

jameshaydon:
> In theory I think all of that logic should be in the CMap datatype (or on
a new datatype that wraps it) - the API that the datatype exposes should be
such that reading and writing tend to increase polling.

I like this idea a lot!

PR #401, opened 2019-01-22
Add notes about packaging

geigerzaehler:
@jkarni, @cloudhead, and me have talked a lot about how to package Radicle. Here is the summary of this discussion.


notes/2018-01-22-packaging.md
```diff
@@ -0,0 +1,104 @@
+Packaging
+=========
+
+This documents explains how we want to package and distribute Radicle.
+
+We will create packages for Brew, Debian, Pacman, and (possibly Nix). This
+packages will be made available through Github Releases.
+
+Haskell binaries will be statically linked with `-O3` optimization.
```
kim: I didn't know `-O3` actually existed, and it is usually advised to use `-O2` over `-O1` **iff** you have **strong empirical evidence** that it doesn't make things worse. So, do we have empirical evidence?

notes/2018-01-22-packaging.md
```diff
@@ -0,0 +1,104 @@
...
+  them and dispatch to them.
+* Radicle apps (implemented as subcommands) need to locate Radicle source files.
+* If Radicle app frontends that are written in radicle the interpreter also needs
+  to locate Radicle source files to be able to import them.
+
+In addition to installed packages the location mechanism needs to work for
+development builds (`stack build`), local tests, and CI tests.
+
+To implement file location of subcommands we are going to use Cabal’s
+[`Paths_radicle`][cabal-paths] module. `getBinDir` will return the folder
```
kim: Please see commercialhaskell/stack#848 -- paths returned by`Paths_*` modules generated by `stack` builds will be absolute paths _on the build machine_, pointing somewhere below `.stack-work/install`

notes/2018-01-22-packaging.md
```diff
@@ -0,0 +1,104 @@
...
+packages the `rad` command will be a script with the following content.
+
+```
+#!/bin/sh
+radicle_bin=/usr/lib/radicle/bin
+RADPATH=/usr/lib/radicle/modules
+exec $radicle_bin/rad "@$"
+```
+
+Setting `radicle_bin` will determine the return value of `getBinDir`. Setting
```
kim: I believe this must be `<executable_name>_bindir`. Still, I think the default path should already point to the "right" directory (ie. build using `cabal`, which allows to tweak those via `cabal configure`).

notes/2018-01-22-packaging.md
```diff
@@ -0,0 +1,104 @@
...
+For Radicle to work two services need to be running: the Radicle daemon and the
+IPFS daemon. For Debian in Pacman we will provide a Systemd Unit file for the
+Radicle daemon as well as installation instructions that include extra steps to
+enable the Radicle daemon service.
+
+The Radicle daemon Systemd Unit has a dependency on the IPFS Systemd Unit. If we
+distribute IPFS with the package we also need to add a Systemd Unit file for
+IPFS.
+
+We will investigate whether Brew provides a similar service management system
```
kim: Yes, launchd. Essentially, one needs to inline a plist. See for example [postgres](https://github.com/Homebrew/homebrew-core/blob/master/Formula/postgresql.rb#L83)

notes/2018-01-22-packaging.md
```diff
@@ -0,0 +1,104 @@
...
+package only minor modifications would need to be made.
+
+1. `stack build -- --flag radicle:release`. The flag disables some targets,
+   enforces static linking and `-O3`.
+2. Move all binaries built with stack to `$DEST/usr/lib/radicle/bin`
+3. Move all Radicle sources files to `$DEST/usr/lib/radicle/modules`
+4. Create `$DEST/usr/bin/rad` wrapper script as mentioned above.
+5. Create a `deb` package using [`fpm`][fpm]. `fpm` takes the Systemd file as a CLI option.
+6. Upload the `deb` file to Github releases
+
```
kim: GPG Signatures?

notes/2018-01-22-packaging.md
kim: Oh and, do we have evidence that all our binaries _can_ be statically linked?

adinapoli-mndc: Yeah, I always used `-O2` myself as well. Interestingly enough, `-O3` is actually not even mentioned in the last [GHC user guide](https://downloads.haskell.org/~ghc/master/users-guide/using-optimisation.html#o-convenient-packages-of-optimisation-flags). Perhaps is more conservative to go with `-O2`? I don't think it would make such a drastic difference anyway.

notes/2018-01-22-packaging.md
geigerzaehler: Any suggestions how to sign the release artifacts?

notes/2018-01-22-packaging.md
geigerzaehler: 👍 Added it to the notes.

notes/2018-01-22-packaging.md
geigerzaehler: I changed the optimization to use `-O2`. Build time should not be too much of a concern since this still runs in only a few minutes on CI.

> Oh and, do we have evidence that all our binaries _can_ be statically linked?

Yes, everything that we want to distribute can be statically linked. (We will not distribute the old server. It uses libpq which cannot not be statically linked)

notes/2018-01-22-packaging.md
kim: If we make releases from a CD pipeline, it boils down to trusting the build machine. Perhaps releases shouldn't be cut from a SaaS provider?

notes/2018-01-22-packaging.md
kim: My question still persists ;)

notes/2018-01-22-packaging.md
geigerzaehler: To circumvent this we set the `radicle_bindir` environment variable. I agree that it would be preferable to use cabal and set this during build time. This would require more work though. I think for now using the environment variable is a good compromise that allows us to implement this quickly. We will change this once we get the cabal build working.

I added all this to the notes with a bit more detail.

notes/2018-01-22-packaging.md
geigerzaehler: I checked and it is is `<package_name>_bindir`. For the second point see my answer above.

notes/2018-01-22-packaging.md
geigerzaehler: > My question still persists ;)

Which question are you referring to? If it is about empirical evidence for `-O2` being better than `-O1`, then no, we don’t. If it is about evidence that the binaries can be statically linked, then I’d say yes. I built them using static linking and was able to run them on at two distributions

notes/2018-01-22-packaging.md
geigerzaehler: I’ll leave this as an open question for now. I added notes in a new section.

PR #407, opened 2019-01-29
Save build cache even on failure

geigerzaehler:
We want to save the build cache for dependencies even if the project code fails to build or the tests fail. Before a test failure would prevent the build cache from being uploaded. This meant that the next build that fixed the tests would need to rebuild all the new dependencies.

### DIFF
--- a/cloudbuild.yaml
+++ b/cloudbuild.yaml
@@ -23,7 +23,30 @@ steps:
       stack config set system-ghc --global true
       stack config set install-ghc --global false
 
-      stack build stylish-haskell hlint weeder
+      stack build \
+        --no-terminal \
+        stylish-haskell hlint weeder
+
+      apt-get -qq update
+      apt-get -yqq install libpq-dev
+
+      stack build \
+        --no-terminal \
+        --dependencies-only
+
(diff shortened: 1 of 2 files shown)


cloudbuild.yaml
```diff
@@ -23,7 +25,19 @@ steps:
...
+
+      apt-get -qq update
+      apt-get -yqq install libpq-dev
+
+      stack build \
+        --no-terminal \
+        --dependencies-only
+
+      source scripts/ci.sh
+      save-cache
```
kim: This does not actually work (but the error is hidden because `|| true`): the step uses the haskell image, which doesn't have `gsutil` installed. The least painful option is probably to just move the "Save cache" step to run after "Build deps", and change the other dependencies accordingly.

kim: Also it seems like there are a couple of opportunities for concurrency:

* "Format", "Lint" and "Build" can run concurrently

   The first step to fail fails the entire build - so these should run _after_ "Save cache".

* "Test" and "Lint (weeder)" can both run concurrently after "Build"

Probably for a separate PR :)

PR #410, opened 2019-01-30
Daemon client and Radicle bindings

geigerzaehler:
Todo
- [x] Start daemon on CI so that tests pass

### DIFF
--- a/exe/RadicleExe.hs
+++ b/exe/RadicleExe.hs
@@ -8,6 +8,7 @@ import           Options.Applicative
 import           System.Directory (doesFileExist)
 
 import           Radicle
+import           Radicle.Daemon.Client (createDaemonClientPrimFns)
 import           Radicle.Internal.Effects (exitCode)
 import           Radicle.Internal.MachineBackend.EvalServer
 import           Radicle.Internal.MachineBackend.Ipfs (ipfsPrimFns)
@@ -89,5 +90,6 @@ opts = Opts
 
 createBindings :: (MonadIO m, ReplM m) => [Text] -> IO (Bindings (PrimFns m))
 createBindings scriptArgs' = do
+    daemonClientPrimFns <- createDaemonClientPrimFns
     evalServerPackendPrimFns <- createEvalServerBackendPrimFns
-    pure $ addPrimFns (replPrimFns scriptArgs' <> evalServerPackendPrimFns <> ipfsPrimFns) pureEnv
+    pure $ addPrimFns (replPrimFns scriptArgs' <> evalServerPackendPrimFns <> ipfsPrimFns <> daemonClientPrimFns) pureEnv
--- a/src/Radicle/Daemon/Client.hs
+++ b/src/Radicle/Daemon/Client.hs
@@ -1,0 +1,100 @@
+-- | Client for the daemons machine API.
+--
+-- Also provides primitive functions to bind the client into the Radicle
+-- interpreter with @daemonClientPrimFns@.
+module Radicle.Daemon.Client
+    ( createDaemonClientPrimFns
+    , newMachine
+    , query
+    , send
+    , MachineId(..)
+    ) where
+
+import           Protolude hiding (TypeError)
+
+import           Control.Monad.Except
+import           GHC.Exts (fromList)
+import qualified Network.HTTP.Client as HttpClient
+import           Servant.Client hiding (Client)
+
+import           Radicle
+import           Radicle.Daemon.HttpApi
+import qualified Radicle.Internal.PrimFns as PrimFns
+
+
+-- | Constraints for all client functions
+type Client m = (MonadIO m, MonadError ServantError m)
+
+-- | Send an input to a machine to be evaluated and applied. Returns the results
+-- of evaluting the inputs.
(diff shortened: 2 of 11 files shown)


src/Radicle/Daemon/Client.hs
```diff
@@ -0,0 +1,99 @@
...
+
+-- * Internal
+
+runClient :: forall a m. (Client m) => ClientM a -> m a
+runClient clientAction = liftEither =<< liftIO runHttp
+  where
+    runHttp :: IO (Either ServantError a)
+    runHttp = do
+        baseUrl <- parseBaseUrl "http://localhost:8909"
+        manager <- HttpClient.newManager HttpClient.defaultManagerSettings
```
jkarni: Managers should be shared rather than created anew for each request.

src/Radicle/Daemon/Client.hs
```diff
@@ -0,0 +1,99 @@
...
+
+
+-- | Constraints for all client functions
+type Client m = (MonadIO m, MonadError ServantError m)
+
+-- | Send an input to a machine to be evaluated and applied. Returns the results
+-- of evaluting the inputs.
+send :: Client m => MachineId -> Seq Value -> m [Value]
+send machineId inputs = do
+    SendResponse {..} <- runClient $ client machineSendEndpoint machineId (SendRequest $ toList inputs)
```
jkarni: Can `runClient`, `result`, and `wrapServantError` be combined into a function `foo` so we get:
```
query :: (Client m) => MachineId -> Value -> Lang m Value
query machineId q = foo $ client machineQueryEndpoint machineId (QueryRequest q)
```
and likewise for `send` and `newMachine`?

src/Radicle/Daemon/Client.hs
```diff
@@ -0,0 +1,99 @@
...
+--
+-- If the Servant error is 'FailureResponse' we wrap the content of the reponse
+-- body. Otherwise we use 'displayException'.
+wrapServantError :: Monad m => ExceptT ServantError m a -> Lang m a
+wrapServantError action = do
+    result <- lift $ runExceptT action
+    case result of
+        Left (FailureResponse response) -> throwErrorHere $ OtherError $ toS $ responseBody response
+        Left e -> throwErrorHere $ OtherError $ toS $ displayException e
+        Right value                  -> pure value
```
jkarni: Weird spacing here.

src/Radicle/Daemon/HttpApi.hs
```diff
@@ -26,57 +33,90 @@ import qualified Paths_radicle as Radicle
...
-singleKey tyName c k = A.withObject tyName $ k >=> pure . c
+instance ToHttpApiData MachineId where
+    toUrlPiece (MachineId id) = id
 
 newtype QueryRequest = QueryRequest { expression :: Value }
   deriving (Generic)
 
 instance A.ToJSON QueryRequest where
-  toJSON = valueToJson . expression
+    toJSON QueryRequest{..} = A.object
```
jkarni: Can we add a [validateEveryToJSON](http://hackage.haskell.org/package/servant-swagger-1.1.7/docs/Servant-Swagger-Test.html#v:validateEveryToJSON) test so we're sure the swagger instances are correct?

exe/RadicleExe.hs
```diff
@@ -90,4 +91,4 @@ opts = Opts
 createBindings :: (MonadIO m, ReplM m) => [Text] -> IO (Bindings (PrimFns m))
 createBindings scriptArgs' = do
     evalServerPackendPrimFns <- createEvalServerBackendPrimFns
-    pure $ addPrimFns (replPrimFns scriptArgs' <> evalServerPackendPrimFns <> ipfsPrimFns) pureEnv
+    pure $ addPrimFns (replPrimFns scriptArgs' <> evalServerPackendPrimFns <> ipfsPrimFns <> daemonClientPrimFns) pureEnv
```
jkarni: Shouldn't we just remove the server stuff?

jkarni: And shouldn't we also remove the `ipfsPrimFns`?

src/Radicle/Daemon/Client.hs
geigerzaehler: `exe/Machines` does not care about the `Lang` monad so an interface with less assumptions prevents unnecessary unwrapping. In general I think using `Client m` rather than `Lang m` makes the interface more flexible.

exe/RadicleExe.hs
jkarni: (As well as all of `MachineBackend/Ipfs.hs`)

geigerzaehler: Yes! But instead of doing it in this PR we should tackle this separately in a cleanup PR.

src/Radicle/Daemon/HttpApi.hs
geigerzaehler: I created a ticket for that.

exe/RadicleExe.hs
jkarni: :+1:. My PR is removing any use of these functions, so I can do it.

PR #411, opened 2019-01-30
Handle my-keys via key-mgmt module & add a key creation command

MeBrei:
Keys are now stored in `~/.config/radicle/my-keys.rad`, where the config directory can be adjusted via `$XDG_CONFIG_HOME`. 

Keys can be created via `rad-create-keys`, if `--force` is appended, an existing `my-keys.rad` is overwritten.

### DIFF
--- a/bin/rad-create-keys
+++ b/bin/rad-create-keys
@@ -1,7 +1,6 @@
 #!/usr/bin/env radicle
 
 (load! (find-module-file! "prelude.rad"))
-(load! (find-module-file! "prelude/cmd-parsing.rad"))
 
 (import prelude/key-management :as 'key-mgmt)
 
--- a/bin/rad-create-keys
+++ b/bin/rad-create-keys
@@ -5,6 +5,18 @@
 
 (import prelude/key-management :as 'key-mgmt)
 
+(def help
+  "rad-create-keys - Radicle Key Management CLI
+
+   Usage:
+        rad-create-keys [--force]
+        rad-create-keys help
+
+   Create a new key pair with `rad-create-keys`. If `--force` is appended, an
+   existing key pair file will be overwritten.
+   Per default, key pairs are stored in `$HOME/.config/radicle/my-keys.rad`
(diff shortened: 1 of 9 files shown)


bin/rad-create-keys
```diff
@@ -0,0 +1,34 @@
+#!/usr/bin/env radicle
+
+(load! (find-module-file! "prelude.rad"))
+(load! (find-module-file! "prelude/cmd-parsing.rad"))
```
jkarni: Can you import cmd-parsing from prelude, and then remove this line?

rad/prelude/key-management.rad
```diff
@@ -0,0 +1,48 @@
...
+ this can be adjusted by setting `$XDG_CONFIG_HOME`."
+ :exports '[read-keys! create-keys! use-fake-keys]}
+
+(import prelude/io :as 'io)
+(import prelude/strings :as 'string)
+(import prelude/lens :unqualified)
+
+(def key-pair-mode (ref {:use-fake-key #f}))
+
+(def use-fake-keys
```
jkarni:
```suggestion
(def use-fake-keys!
```

rad/prelude/key-management.rad
```diff
@@ -0,0 +1,48 @@
...
+  "Bypass reading the keys from `my-keys.rad`. This is intended for testing."
+  (fn []
+    (set-ref key-pair-mode (@ :use-fake-key) #t)
+    (set-ref key-pair-mode (@ :fake-key) (gen-key-pair! (default-ecc-curve)))))
+
+(def base-path
+  (fn []
+    (def xgd-config-home (string/unlines (io/shell-with-stdout! "echo $XDG_CONFIG_HOME" "")))
+    (def home (string/unlines (io/shell-with-stdout! "echo $XDG_CONFIG_HOME" "")))
```
jkarni: I think this is supposed to be `$HOME`?

bin/rad-create-keys
MeBrei: I realised that I am actually not using cmd parsing here, so I removed it.

PR #415, opened 2019-02-01
Integrate daemon

jkarni:
This switches all of our machines to IPFS.

It also renames a lot of "chain" to "machine", and removes `:enter-chain`.


jkarni:
Got a few lingering chain.rad to fix.

src/Radicle/Internal.hs
```diff
@@ -13,3 +16,8 @@ import           Radicle.Internal.Parse as X
 import           Radicle.Internal.Pretty as X
 import           Radicle.Internal.PrimFns as X
 import           Radicle.Internal.Type as X
+
+createBindings :: (MonadIO m, ReplM m) => [Text] -> IO (Bindings (PrimFns m))
```
geigerzaehler: Maybe we should call this `createReplBindings` or `createImpureBindings`. Please also add documentation.

exe/ReferenceDoc.hs
```diff
@@ -33,13 +34,15 @@ instance FromJSON Content
 main :: IO ()
 main = do
     content <- decodeFileThrow "reference-doc.yaml"
+    daemonClientPrimFns <- createDaemonClientPrimFns
+    let bindings = addPrimFns (replPrimFns [] <> daemonClientPrimFns) pureEnv
```
geigerzaehler: Can’t we use `createBindings` from `Radicle.Internal`?

exe/Doc.hs
```diff
@@ -26,7 +27,9 @@ run f = do
     let extensions = def { readerExtensions = githubMarkdownExtensions }
     pand <- runIOorExplode $ readMarkdown extensions txt
     let code = T.append "(load! (find-module-file! \"prelude.rad\"))" (getCode pand)
-    res <- runLang (replBindings []) $ interpretMany (toS f) $ code
+    daemonClientPrimFns <- createDaemonClientPrimFns
+    let bindings = addPrimFns (replPrimFns [] <> daemonClientPrimFns) pureEnv
```
geigerzaehler: Can’t we use `createBindings` from `Radicle.Internal`?

PR #417, opened 2019-02-01
Only change state to accepted, if all git commands are through

MeBrei:
On accepting a diff, the state will only be changed once all git commands were done.

This still doesn't handle failed git commands properly (additional user feedback to resolve the problem & not all git commands are wrapped in error handling), but this should be addressed after some git refactoring.

### DIFF
--- a/bin/rad-diff
+++ b/bin/rad-diff
@@ -284,11 +284,10 @@
               (put-str! message)))
           (verify-diff-number n chain-ref)
           (if (eq? new-state :accepted)
-            (do
-              (match (push-diff! chain-ref n)
-                :ok (edit-state-with-feedback!
-                      (string-append "Diff " (show n) " has been " (pretty-state new-state) " and merged."))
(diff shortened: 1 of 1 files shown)


bin/rad-diff
```diff
@@ -277,14 +276,21 @@
...
+              (simple-edit-diff! chain-ref n {:state new-state})
+              (put-str! message)))
           (verify-diff-number n chain-ref)
-          (simple-edit-diff! chain-ref n {:state new-state})
           (if (eq? new-state :accepted)
             (do
-              (push-diff! chain-ref n)
-              (put-str! (string-append "Diff " (show n) " has been " (pretty-state new-state) " and merged.")))
-            (put-str! (string-append "Diff " (show n) " has been " (pretty-state new-state)))))
+              (match (push-diff! chain-ref n)
```
geigerzaehler: I think the `do` above is redundant now

PR #418, opened 2019-02-01
rad-create-keys: Feedback when successfully created

geigerzaehler:
`rad-create-keys` now outputs something to the user when the key file is successfully created.


MeBrei:
Nice addition!
In the case of `rad-create-keys` the user feedback might not be completely correct, as it might overwrite an already existing file. But this could also be a later adjustment.

PR #420, opened 2019-02-06
Daemon doesn't fail if it can't init a reader-mode machine

jameshaydon:


jkarni:
Should we build images for the daemon now, and try getting the tests to work with that instead? I imagine we'd want to be deploying daemons quite often.
(cc @geigerzaehler )

exe/Daemon.hs
```diff
@@ -515,31 +527,32 @@ writeInputs id is nonce = do
...
+                  in if more' > 0
+                     then (True, HighFreq more')
+                     else (False, LowFreq)
+                -- Low frequency polling is every 10 seconds.
+                LowFreq -> (delta > lowFreqPollPeriod, LowFreq)
+        when shouldPoll $ do
+          logInfo Debug "Polling.." [("machine-id", getMachineId machineId)]
+          refreshAsReader machineId >> pure ()
+        modifyMachine machineId $ \m' -> pure (m' { machinePolling = newPoll }, ())
+      _ -> pure () -- Uninitialised readers are (for the moment) ignored.
```
jkarni: Why not (low-freq) poll them as well?

cloudbuild.yaml
```diff
@@ -202,4 +163,3 @@ steps:
       stack exec -- radicle-daemon &
 
       stack test :e2e
-      stack exec -- radicle test/machine-backends.rad radicle-server
```
geigerzaehler: Please also remove the corresponding file `test/machine-backends.rad`. Otherwise we’ll forget about it

geigerzaehler:
> Should we build images for the daemon now, and try getting the tests to work with that instead? I imagine we'd want to be deploying daemons quite often.

Seems likely that we want to build an image for the daemon in the future. But right now I see no immediate advantage to building and running it with docker.

cloudbuild.yaml
jameshaydon: :+1:

exe/Daemon.hs
jameshaydon: :+1:

src/Radicle/Daemon/Common.hs
```diff
@@ -26,18 +27,19 @@ data Polling = HighFreq Int64 | LowFreq
...
+    , machineState        :: Bindings (PrimFns Identity)
+    , machineLastIndex    :: Maybe MachineEntryIndex
+    , machineMode         :: ReaderOrWriter
+    , machineSubscription :: TopicSubscription
+    , machineLastUpdated  :: Time.SystemTime
+    , machinePolling      :: Polling
     } deriving (Generic)
 
-newtype Chains id idx subs = Chains { getChains :: CMap.CMap id (Chain id idx subs) }
+data CachedMachine = UninitialisedReader Time.SystemTime | Cached Machine
```
jkarni: Can you put a comment saying what this `SystemTime` is for/about?

jameshaydon: :+1:

PR #424, opened 2019-02-07
Remove seen/unseen functionality from diff app

MeBrei:


bin/rad-diff
```diff
@@ -207,31 +199,22 @@
...
         _       (fn [x] (elem? (lookup :state (lookup :diff x)) options))))
 
   (def items
     (map mk-key
       (filter filt
-        (match [(elem? :new options) (read-file-key! ".radicle-diffs" :last-mark-done)]
-           [#t [:just 't]] (recent-activity chain-ref t)
-           _               (map (fn [i] {:diff i})
-                                (reverse (values (list-diffs chain-ref))))))))
+        (map (fn [i] {:diff i}) (reverse (values (list-diffs chain-ref)))))))
```
geigerzaehler: I think we can remove the wrapping of the diff in `{:diff i}`. This would eliminate all the `:diff` lookups and matches against `diff`.

bin/rad-diff
```diff
@@ -207,41 +199,32 @@
...
       (fn [v]
         (match v
-          (/as 'a {:activity :new-diff
-                   :diff     'diff})
-            [(string-append "[new/changed] " (pretty-headline diff)) a]
-          (/as 'a {:activity :new-comments
-                   :diff     'diff})
-            [(string-append "[new comments] " (pretty-headline diff)) a]
-          (/as 'i {:diff 'diff})
+          (/as 'i 'diff)
```
geigerzaehler: Now you can remove the `match` clause completely
~~~clojure
(fn [diff] [(pretty-headline diff) diff)])
~~~

PR #426, opened 2019-02-07
Handle failures in project checkout

MeBrei:
This handles if the directory already exists and if cloning the repository fails.

Refactoring processing of git commands will happen in a different PR, which will make this code cleaner.

### DIFF
--- a/bin/rad-project
+++ b/bin/rad-project
@@ -60,8 +60,8 @@
     (match (process! "test" ["-d" name] "")
       :ok (do
             (put-str! (string-append
-                        "The directory \"" name "\" already exists. "
-                        "It needs to be removed to checkout the project in this directory."))
+                        "The directory \"" name "\" already exists.\n"
+                        "Refusing to continue."))
             (exit! 1))
       _   (put-str! name))
     (match (process! "git" ["clone" origin name] "")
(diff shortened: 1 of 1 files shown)


bin/rad-project
```diff
@@ -57,9 +57,20 @@
...
     (def name (lookup :name (get-meta! rsm)))
     (def origin (lookup :id (first-rsm-of-type! rsm :rad-repo)))
-    (put-str! name)
-    (process! "git" ["clone" origin name] "")
-    (cd! name)
+    (match (process! "test" ["-d" name] "")
+      :ok (do
+            (put-str! (string-append
+                        "The directory \"" name "\" already exists. "
+                        "It needs to be removed to checkout the project in this directory."))
```
jkarni: Removing an existing dir is dangerous advice to give :) Maybe just "Refusing to continue."

bin/rad-project
```diff
@@ -57,9 +57,20 @@
...
+            (exit! 1))
+      _   (put-str! name))
+    (match (process! "git" ["clone" origin name] "")
+      :ok (cd! name)
+      _   (do
+            (put-str! (string-append
+                        "The repo is currently not available. "
+                        "Cloning \"" name "\" from \"" origin "\" failed."))
```
jkarni: Maybe add something like `"This may be because you have no internet connection, or the IPNS link is stale, or because no online node with the data could be found."`

PR #428, opened 2019-02-07
Daemon pins all expressions in reader mode.

jameshaydon:


src/Radicle/Ipfs.hs
```diff
@@ -188,10 +189,21 @@ instance FromJSON DagPutResponse where
...
+      Left _     -> fail "invalid CID"
+      Right cids -> pure $ PinResponse cids
+
+-- | Get and pin a DAG node.
 dagGet :: FromJSON a => Address -> IO a
 dagGet addr = do
-    result <- ipfsHttpGet "dag/get" [("arg", addressToText addr)]
+    let params = [("arg", addressToText addr)]
+    result <- ipfsHttpGet "dag/get" params
+    PinResponse _ <- ipfsHttpGet "pin/add" params
```
geigerzaehler: I would prefer to introduce a seperate `pinAdd` method. The `Radicle.Ipfs` module should mirror the IPFS API. It is then up to the application logic to pin a machine explicitly

jkarni: Unless you add `progress=false`, this `PinResponse` datatype is already not really a perfect match with the response, so I'm wondering why have it at all (instead of just doing `() <- ipfsHttpGet ...`

(No strong opinions though)

jameshaydon: Progress is only returned if `progress=true`.

PR #434, opened 2019-02-12
Refactor git and handle git command failures

MeBrei:
This handles most of the git commands. The error messages might not be comprehensive.

Some unhandled cmds remain in `new-or-forked-project!` in `rad-project`, but they are also part of open TODOs.

### DIFF
--- a/rad/prelude/io-utils.rad
+++ b/rad/prelude/io-utils.rad
@@ -1,6 +1,7 @@
 {:module 'prelude/io-utils
  :doc "IO-related utilities"
- :exports '[fzf-select! edit-in-editor! get-git-config! set-git-config!]}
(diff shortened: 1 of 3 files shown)


rad/prelude/io-utils.rad
```diff
@@ -45,7 +46,32 @@
...
+                (put-str! (string-append
+                            "git config user.name is not set. "
+                            "Set the user name and rerun you command.")))
+      'name  name)))
+
+(def get-git-commit-data!
+  "Get data from a `commit` via `show` specified by `format`"
+  (fn [format commit]
+    (unlines
+      (process-with-stdout! "git" ["show" "-s" (string-append "--format=" format) commit] ""))))
```
jkarni: This should probably throw an exception if the process exits with an error.

PR #437, opened 2019-02-13
Apps catch daemon errors.

jameshaydon:
- [x] Use an app without the Radicle daemon running
- [x] Use an app with the Radicle daemon but not IFPS running
- [x] Use an app without internet
- [x] Use a non-exsiting machine ID
- [x] Use a machine where the IPNS name resolves but the data cannot be reached
- [x] What happens when a send does not receive an ack in time (might need to tweak daemon code to simulate)
- [x] cloning a non existing project

### DIFF
--- a/bin/rad-diff
+++ b/bin/rad-diff
@@ -296,18 +296,22 @@
     (/cmd-opts "list" opts cmd-options help)))
 
 (def args (get-args!))
-(match args
-  (/list-cmd 'options) (list-items (machine-name) options)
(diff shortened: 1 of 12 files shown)


jameshaydon:
@MeBrei 
> Only thing I noticed is that the messages for IPFS related errors are less descriptive ie are not hinting to any possible reason or ways to fix it.

The IPFS errors come directly from the IPFS daemon, but yeah I agree they can be a little cryptic. It's still better to keep them verbatim to improve googleability, but I agree we should add some hints as to how to resolve the issues as a next step.

PR #439, opened 2019-02-13
Use new list filter syntax

MeBrei:
This adds the new filter syntax `--filter-by-state`, `--state`, `-s` for both apps and moves the fzf functionality of the issue app to the option `--fancy`

### DIFF
--- a/bin/rad-diff
+++ b/bin/rad-diff
@@ -5,23 +5,19 @@
 (load! (find-module-file! "prelude/io-utils.rad"))
 (load! (find-module-file! "monadic/project.rad"))
 
-(import prelude/cmd-parsing :unqualified)
 (import prelude/validation :as 'validation)
 
 (def machine-name (fn [] (get-rsm-of-type! :rad-diff)))
 
-(def valid-list-options
-  ["--pending" "--retracted" "--accepted" "--rejected" "--fancy"])
-
-(def list-opts-help
-  (foldl (fn [a opt] (string-append a " [" opt "]")) "" valid-list-options))
+(def valid-states
+  ["pending" "retracted" "accepted" "rejected"])
 
 (def help
   (string-append
   "rad-diff - Radicle Diff CLI
 
    Usage:
-        rad-diff list " list-opts-help "
+        rad-diff list [--filter-by-state | --state | -s <state>] [--fancy]
         rad-diff propose <commit>
         rad-diff [retract|accept|reject|show] <diff-no>
         rad-diff comment <diff-no> <comment>
@@ -191,9 +187,9 @@
       (fn [diff] [(pretty-headline diff) diff]))
 
     (def filt
-      (match (filter (fn [opt] (not (eq? opt :fancy))) options)
+      (match (lookup :state options)
         []      (fn [x] #t)
-        _       (fn [x] (elem? (lookup :state x) options))))
+        'args   (fn [x] (elem? (lookup :state x) args))))
 
   (def items
     (map mk-key
@@ -203,7 +199,7 @@
     (def items-dict (dict-from-seq items))
     (if (empty-seq? items)
       (put-str! "No diffs!")
-      (if (elem? :fancy options)
+      (if (lookup :fancy options)
         (match (fzf-select-with-preview!
                  (map first items)
                  (fn [s] (pretty-diff (lookup s items-dict))))
@@ -271,16 +267,22 @@
         (cmd-parse-failure "Diff numbers must be whole numbers!"))
     (cmd-parse-failure "To create a comment, please provide the diff number."))))
 
-(def list-options
-  (fn [os]
-    (map (fn [o]
-             (if (member? o valid-list-options)
-               (read (string-append ":" (drop 2 o)))
-               (cmd-parse-failure (string-append "Invalid list option: " o)))) os)))
+(def opts-match
+ (fn [opt]
+   (match opt
+     "--filter-by-state" :state
+     "--state"           :state
+     "-s"                :state
+     "--fancy"           :fancy
+     'invalid            (parse-failure (string-append "Invalid option " invalid " for command list.") help))))
+
+(def opt-args {:state valid-states})
+
+(def opts-default {:state [] :fancy #f})
 
 (def /list-cmd
   (fn [opts]
-    (/cmd-1-opts "list" opts list-options)))
+    (/cmd-opts "list" opts opt-args opts-match opts-default help)))
 
 (def args (get-args!))
 (match args
--- a/bin/rad-diff
+++ b/bin/rad-diff
@@ -12,15 +12,13 @@
 (def valid-states
   ["pending" "retracted" "accepted" "rejected"])
 
-(def status-opts ["--filter-by-state" "--state" "-s"])
+(def status-opts ["-s" "--state" "--filter-by-state" ])
 
 (def list-opts-help
   (string-append
     "[("
     (string/intercalate " | " status-opts)
-    ") ("
-    (string/intercalate " | " valid-states)
-    ")] [--fancy]"))
+    ") <state>] [--fancy]"))
 
 (def help
   (string-append
@@ -29,7 +27,7 @@
    Usage:
         rad-diff list " list-opts-help "
         rad-diff propose <commit>
-        rad-diff [retract|accept|reject|show] <diff-no>
+        rad-diff [retract | accept | reject | show] <diff-no>
         rad-diff comment <diff-no> <comment>
         rad-diff help
 
--- a/bin/rad-diff
+++ b/bin/rad-diff
@@ -12,12 +12,22 @@
 (def valid-states
   ["pending" "retracted" "accepted" "rejected"])
 
+(def status-opts ["--filter-by-state" "--state" "-s"])
+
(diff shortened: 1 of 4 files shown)


bin/rad-diff
```diff
@@ -5,23 +5,19 @@
...
+(def valid-states
+  ["pending" "retracted" "accepted" "rejected"])
 
 (def help
   (string-append
   "rad-diff - Radicle Diff CLI
 
    Usage:
-        rad-diff list " list-opts-help "
+        rad-diff list [--filter-by-state | --state | -s <state>] [--fancy]
```
geigerzaehler: This is a bit confusing. It looks like `--state` and `--filter-by-state` do not take arguments. We should also document what the possible states are.

rad/prelude/cmd-parsing.rad
```diff
@@ -27,32 +29,57 @@
...
+      (def default {:state [] :fancy #f})
+      (def opt-match
+        (fn [opt]
+          (match opt
+            "--filter-by-state" :state
+            "--state"           :state
+            "-s"                :state
+            "--fancy"           :fancy
+            'invalid            (parse-failure (string-append "Invalid option " invalid " for command list.") help)))))]
+  [((/cmd-opts "list" 'opts multi-opts opt-match default help) ["list"] ) ==> [:just {opts {:state [] :fancy #f}}]]
```
geigerzaehler: Would it make sense to extract the repeated `(/cmd-opts "list" 'opts multi-opts opt-match default help)`?

rad/prelude/cmd-parsing.rad
```diff
@@ -27,32 +29,57 @@
...
+      (def multi-opts {:state ["acc" "prop"]})
+      (def default {:state [] :fancy #f})
+      (def opt-match
+        (fn [opt]
+          (match opt
+            "--filter-by-state" :state
+            "--state"           :state
+            "-s"                :state
+            "--fancy"           :fancy
+            'invalid            (parse-failure (string-append "Invalid option " invalid " for command list.") help)))))]
```
geigerzaehler: Instead of `opt-match` being a function it could be a dictionary. The user of `/cmd-opts` would not need to provide the `'invalid` case every time.

rad/prelude/cmd-parsing.rad
```diff
@@ -27,32 +29,57 @@
...
-  combination of options that are allowed by opts-map"
-  (fn [cmd opts opts-map]
+(def /cmd-opts
+  "Matches the command `cmd` and a combination of optional parameters, where
+  `opt-match` matches the options to the keys of `multi-ops`. `multi-ops` is a
+  dict of option-keys with multiple arguments. If no options are passed, the
+  `default` is returned."
+  (fn [cmd opts multi-opts opt-match default help]
+    (def-rec parse-cli
+      (fn [args default]
```
geigerzaehler: You don’t need to pass `default`. The value is always available in the closure from the arguments to `/cmd-opts`.

rad/prelude/cmd-parsing.rad
```diff
@@ -27,32 +29,57 @@
...
+        (def parse-arg
+          (fn [name args]
+            (match args
+              (/cons 'val 'rest)
+                  (if (elem? val (lookup name multi-opts))
+                    (modify-map name (fn [x] (cons (read (string-append ":" val)) x)) (parse-cli rest default))
```
geigerzaehler: I would prefer to have the caller of `/cmd-opts` do the conversion from string to symbol. This would make the code less complex and more flexibel when we want to reuse it.

rad/prelude/cmd-parsing.rad
```diff
@@ -27,32 +29,57 @@
...
-(def /cmd-1-opts
-  "Matches a command `cmd` with one argument and an optional
-  combination of options that are allowed by opts-map"
-  (fn [cmd opts opts-map]
+(def /cmd-opts
+  "Matches the command `cmd` and a combination of optional parameters, where
+  `opt-match` matches the options to the keys of `multi-ops`. `multi-ops` is a
+  dict of option-keys with multiple arguments. If no options are passed, the
+  `default` is returned."
+  (fn [cmd opts multi-opts opt-match default help]
```
geigerzaehler: What do you think about the following alternative API for `/cmd-opts`. At the moment we have an argument for every attribute of an option where the argument maps option names to attribute. Instead I’d suggest having only a list of option specifiers as an argument. For example this would look like this
~~~clojure
(/cmd-opts "list" [
    { :key :state :type :multi-option :options ["-s", "--state", "--filter-by-state"]  :possible-values ["acc", "closed"] }
    { :key :fancy :type :flag :options ["--fancy"] :default #f }
    ])
~~~

bin/rad-diff
MeBrei: Yes, I thought the same, but wasn't sure how to fit all this info into the section.
Would this work? `rad-diff list [(--filter-by-state | --state | -s) (proposed | retracted | accepted | rejected)]`

geigerzaehler: Yes, that would work. Maybe @juliendonck wants to chime in.

rad/prelude/cmd-parsing.rad
```diff
@@ -27,32 +30,60 @@
...
+    (def-rec parse-cli
+      (fn [args]
+        (def parse-arg
+          (fn [opt args]
+            (def opt-key (lookup :key opt))
+            (match args
+              (/cons 'val 'rest)
+                  (if (elem? val (lookup :possible-values opt))
+                    (modify-map opt-key (fn [x] (cons val x)) (parse-cli rest))
+                    (parse-failure (string-append "Invalid argument " val " for option " (show opt-key) ".") help))
```
geigerzaehler: At the moment `--state --fancy` warns you that `--fancy` is an invalid value for `--state`. It should warn you that `--state` is missing an argument. I think this can be remedied by checking whether `val` starts with `-`.

rad/prelude/cmd-parsing.rad
```diff
@@ -27,32 +30,60 @@
...
+        (map (fn [opt] (list (lookup :key opt) (lookup :default opt))) cmd-opts)))
+    (def-rec parse-cli
+      (fn [args]
+        (def parse-arg
+          (fn [opt args]
+            (def opt-key (lookup :key opt))
+            (match args
+              (/cons 'val 'rest)
+                  (if (elem? val (lookup :possible-values opt))
+                    (modify-map opt-key (fn [x] (cons val x)) (parse-cli rest))
```
geigerzaehler: We should only do this if the type is `:multi-opt`. Otherwise we should just update the value.

rad/prelude/cmd-parsing.rad
```diff
@@ -27,32 +30,60 @@
...
 
-(def /cmd-1-opts
-  "Matches a command `cmd` with one argument and an optional
-  combination of options that are allowed by opts-map"
-  (fn [cmd opts opts-map]
+(def /cmd-opts
+  "Matches the command `cmd` and a combination of optional parameters, where
+  `opt-match` matches the options to the keys of `multi-ops`. `multi-ops` is a
+  dict of option-keys with multiple arguments. If no options are passed, the
+  `default` is returned."
```
geigerzaehler: Could you update the documentation?

rad/prelude/cmd-parsing.rad
```diff
@@ -27,32 +30,60 @@
...
+        [
+          { :key :state :type :multi-opt :options ["-s" "--state" "--filter-by-state"] :possible-values ["acc" "prop"] :default []}
+          { :key :fancy :type :flag :options ["--fancy"] :default #f }
+        ])
+      (def /cmd (/cmd-opts "list" 'opts cmd-opts help)))]
+  [(/cmd ["list"] ) ==> [:just {opts {:state [] :fancy #f}}]]
+  [(/cmd ["list" "--fancy"] ) ==> [:just {opts {:state [] :fancy #t}}]]
+  [(/cmd ["list" "--state" "acc"] ) ==> [:just {opts {:state ["acc"] :fancy #f}}]]
+  [(/cmd ["list" "--fancy" "-s" "acc"] ) ==> [:just {opts {:state ["acc"] :fancy #t}}]]
+  [(/cmd ["list" "--state" "prop" "-s" "acc"] ) ==> [:just {opts {:state ["prop" "acc"] :fancy #f}}]])
```
geigerzaehler: Could you add test cases for the parse failures? We should test an invalid option, a missing argument, an invalid option argument, and a positional argument (i.e `(/cmd ["list" "arg"])`.

rad/prelude/cmd-parsing.rad
MeBrei: Do you mean you want to prepare the code for a possible third option where an option requires only a single value? I am not sure if it makes sense to include more option types here now without having explicit use for them, as there could be a variety of different option types.

geigerzaehler: Good point, YAGNI. Let’s keep it like it is for now

bin/rad-diff
MeBrei: I talked with @juliendonck and we decided to not list all states in the help text for simplicity. Instead the error message for invalid state arguments will return a list of valid states.

PR #443, opened 2019-02-13
Consolidate IPFS errors

jameshaydon:
Waiting on #437.

### DIFF
--- a/bin/rad-project
+++ b/bin/rad-project
@@ -56,22 +56,21 @@
     (def name (lookup :name (get-meta! project)))
     (def origin (lookup :id (first-rsm-of-type! project :rad-repo)))
     (match (process! "test" ["-d" name] "")
-           :ok (do
-                 (put-str! (string-append
-                            "The directory \"" name "\" already exists.\n"
-                            "Refusing to continue."))
-                 (exit! 1))
-           _   (put-str! name))
-    (match (process! "git" ["clone" origin name] "")
-           :ok (cd! name)
-           _   (do
-                 (put-str! (string-append
-                            "The repo is currently not available. "
-                            "Cloning \"" name "\" from \"" origin "\" failed.\n"
-                            "This may be because you have no internet connection, "
-                            "or the IPNS link is stale, or because no online node "
-                            "with the data could be found."))
-                 (exit! 1)))
+      :ok (do
+            (put-str! (string-append
+                        "The directory \"" name "\" already exists.\n"
+                        "Refusing to continue."))
+            (exit! 1))
+      _   (put-str! name))
+    (process-git-with-exit!
+      ["clone" origin name]
+      (string-append
+        "The repo is currently not available. "
+        "Cloning \"" name "\" from \"" origin "\" failed.\n"
+        "This may be because you have no internet connection, "
+        "or the IPNS link is stale, or because no online node "
+        "with the data could be found."))
+    (cd! name)
     (set-git-config! "radicle.project-id" project-url)))
 
 (def new-project!
--- a/bin/rad-project
+++ b/bin/rad-project
@@ -56,22 +56,22 @@
     (def name (lookup :name (get-meta! project)))
     (def origin (lookup :id (first-rsm-of-type! project :rad-repo)))
     (match (process! "test" ["-d" name] "")
-      :ok (do
-            (put-str! (string-append
-                        "The directory \"" name "\" already exists.\n"
-                        "Refusing to continue."))
-            (exit! 1))
-      _   (put-str! name))
+           :ok (do
+                 (put-str! (string-append
+                            "The directory \"" name "\" already exists.\n"
+                            "Refusing to continue."))
+                 (exit! 1))
+           _   (put-str! name))
     (match (process! "git" ["clone" origin name] "")
-      :ok (cd! name)
-      _   (do
-            (put-str! (string-append
-                        "The repo is currently not available. "
-                        "Cloning \"" name "\" from \"" origin "\" failed.\n"
-                        "This may be because you have no internet connection, "
-                        "or the IPNS link is stale, or because no online node "
-                        "with the data could be found."))
-            (exit! 1)))
+           :ok (cd! name)
+           _   (do
+                 (put-str! (string-append
+                            "The repo is currently not available. "
+                            "Cloning \"" name "\" from \"" origin "\" failed.\n"
+                            "This may be because you have no internet connection, "
+                            "or the IPNS link is stale, or because no online node "
+                            "with the data could be found."))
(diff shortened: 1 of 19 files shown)


src/Radicle/Ipfs.hs
```diff
@@ -163,12 +184,12 @@ subscribe topic messageHandler = runResourceT $ do
...
 publish :: Text -> LByteString -> IO ()
 publish topic message =
-    void $ ipfsHttpPost' "pubsub/pub" [("arg", topic)] "data" message
+    catchHttp $ void $ ipfsHttpPost' "pubsub/pub" [("arg", topic)] "data" message
 
 newtype KeyGenResponse = KeyGenResponse IpnsId
 
 keyGen :: Text -> IO KeyGenResponse
-keyGen name = ipfsHttpGet "key/gen" [("arg", name), ("type", "ed25519")]
+keyGen name = catchHttp $ ipfsHttpGet "key/gen" [("arg", name), ("type", "ed25519")]
```
geigerzaehler: Can’t we move `catchHttp` to `ipfsHttpGet` and `ipfsHttpPost`?

src/Radicle/Ipfs.hs
```diff
@@ -38,33 +38,54 @@ import           Data.Conduit ((.|))
...
+      IpfsException msg -> "ipfs: " <> toS msg
+      IpfsExceptionNoDaemon -> "ipfs: Daemon not reachable, run 'rad ipfs daemon'"
+      IpfsExceptionInvalidResponse url _ -> "ipfs: Cannot parse IPFS daemon response for " <> toS url
+      IpfsExceptionTimeout -> "ipfs: Daemon took too long to respond"
+      IpfsExceptionErrResp msg -> "ipfs: " <> toS msg
+      IpfsExceptionErrRespNoMsg -> "ipfs: filed with no error message"
+
+-- | Turn HTTP exceptions into 'IpfsException's.
+catchHttp :: forall a. IO a -> IO a
+catchHttp io = catch io httpHdlr
```
geigerzaehler: The name `catchHttp` seems a bit confusing to me. Maybe `mapHttpException`?

src/Radicle/Ipfs.hs
```diff
@@ -38,33 +38,54 @@ import           Data.Conduit ((.|))
...
 
 instance Exception IpfsException where
-    displayException (IpfsException msg) = "ipfs: " <> toS msg
-    displayException IpfsExceptionNoDaemon = "ipfs: Daemon not reachable, run 'rad ipfs daemon"
-    displayException (IpfsExceptionInvalidResponse url _) = "ipfs: Cannot parse IPFS daemon response for " <> toS url
-
-
+    displayException = \case
+      IpfsException msg -> "ipfs: " <> toS msg
+      IpfsExceptionNoDaemon -> "ipfs: Daemon not reachable, run 'rad ipfs daemon'"
```
geigerzaehler: The command is not correct anymore. Since most people are going to use systemd to run the deamon this might also be confusing. Maybe we could replace the message with "Cannot connect to Radicle IPFS daemon".

src/Radicle/Ipfs.hs
jameshaydon: :+1:

exe/Daemon.hs
```diff
@@ -202,25 +206,25 @@ send id (Api.SendRequest expressions) = do
       let isResponse = \case
             New InputsApplied{ nonce = Just nonce' } | nonce' == nonce -> True
             _ -> False
-      asyncMsg <- liftIO $ async $ subscribeOne sub ackWaitTime isResponse (logNonDecodableMsg id)
-      ipfs id $ publish id (Submit SubmitInputs{..})
+      asyncMsg <- machineIpfs id $ async $ subscribeOne sub ackWaitTime isResponse (logNonDecodableMsg id)
```
geigerzaehler: Do we need `machineIpfs id` here? It might be confusing because this will never throw an `IpfsException`. `liftIO` should suffice

rad/prelude/machine.rad
```diff
@@ -195,4 +196,11 @@ Do `rad-create-keys` to create your own key pair and rerun your command.")
...
   (fn [machine machine-id cmd payload]
-  (send! machine [(list cmd (sign-entity! payload machine-id))])))
+    (send! machine [(list cmd (sign-entity! payload machine-id))])))
+
+(def catch-daemon!
+  "Catches all `radicle-daemon` related errors and just prints them out to the
+  user."
+  (fn [f]
+    (catch 'daemon-error (f)
+      (fn [e] (put-str! (lookup :info e))))))
```
geigerzaehler: Shouldn’t we also exit with a non-zero exit code?

bin/rad-project
```diff
@@ -56,22 +56,22 @@
...
-                        "The directory \"" name "\" already exists.\n"
-                        "Refusing to continue."))
-            (exit! 1))
-      _   (put-str! name))
+           :ok (do
+                 (put-str! (string-append
+                            "The directory \"" name "\" already exists.\n"
+                            "Refusing to continue."))
+                 (exit! 1))
+           _   (put-str! name))
```
geigerzaehler: These whitespace changes are a bit confusing. We should avoid them in the future if possible to keep the review size small

src/Radicle/Daemon/Client.hs
```diff
@@ -111,6 +111,8 @@ wrapServantError :: Monad m => ExceptT Servant.ServantError m a -> Lang m a
 wrapServantError action = do
     result <- lift $ runExceptT action
     case result of
-        Left (Servant.FailureResponse response) -> throwErrorHere $ OtherError $ toS $ Servant.responseBody response
-        Left e -> throwErrorHere $ OtherError $ toS $ displayException e
+        Left (Servant.FailureResponse response) -> throwErrorHere $ DaemonError $ toS $ Servant.responseBody response
+        Left e -> case e of
+          Servant.ConnectionError _ -> throwErrorHere $ DaemonError "Could not connect to radicle-daemon. Most likely it is not running. Please start it with 'radicle-daemon'."
```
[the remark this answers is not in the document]
jameshaydon: Hm, I think we should have the hint here, this is the error that users using apps will see. But I will say to checks docs to see how to start it instead.

bin/rad-project
jameshaydon: Well, I think we should rather have a standard formatting.

MeBrei:
> This is a great improvement. Only thing I noticed is that the messages for IPFS related errors are less descriptive ie are not hinting to any possible reason or ways to fix it. The radicle-daemon error in contrast suggests starting the daemon.

This comment was meant for #437 . Added the comment there as well

PR #446, opened 2019-02-14
Show diff list in a table format

MeBrei:
This formats the output of `rad-diff list` to a table. The width of the columns are hardcoded and content that exceeds the limits are cut off. 

An alternative could be to get the maximum length of each column first and then apply these to each cell, but this would mean going through all diffs twice and might end up slowing it down.

### DIFF
--- a/bin/rad-diff
+++ b/bin/rad-diff
@@ -155,6 +155,32 @@
   (fn [state]
     (drop 1 (show state))))
 
+(def cells
+  {:state       10
+   :number      10
+   :title       50
+   :author      20
+   :modified-at 20})
+
+(def wrap-cell
+  (fn [phrase cell-key]
+    (def cell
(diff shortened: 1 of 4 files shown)


bin/rad-diff
```diff
@@ -153,16 +153,46 @@
...
     (drop 1 (show state))))
 
+(def cells
+  {:state       10
+   :number      10
+   :title       50
+   :author      20
+   :modified-at 20})
+
+(def wrap-cell
```
jkarni: Can you move 
a) abstract a `pad-to` from this (that takes a str and a length and fills it with spaces or cuts it out)
b) move it to `prelude/strings` 
c) and write a couple of tests?

MeBrei: Due to some weird build failures I had to revert the changes extracting the function. It has to be done at a later time, when we know where the problems came from. (For further investigation check out https://github.com/oscoin/radicle/pull/458)

PR #454, opened 2019-02-15
Replicate app

jkarni:
Introduces `rad-replicate`, which makes replicating to a public daemon easier.

We don't yet have a public daemon, so I tested this with the local one.

### DIFF
--- a/bin/rad-replicate
+++ b/bin/rad-replicate
@@ -1,0 +1,51 @@
+#!/usr/bin/env radicle
+
+(load! (find-module-file! "monadic/project.rad"))
+
+(def help
+  "rad-replicate - Replicate radicle projects
+
+   Usage:
+        rad-replicate [URL]
+  ")
+
+(def default-url "")
+
+(def get-machines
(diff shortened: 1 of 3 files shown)


bin/rad-replicate
```diff
@@ -0,0 +1,73 @@
...
+
+   Usage:
+        rad-replicate [URL]
+
+  If not provided, the URL defaults to http://replicate.radicle.xyz
+  ")
+
+(def default-url
+  "Default replication URL"
+  "http://replicate.radicle.xyz")
```
MeBrei: You could use `default-url` in the help text, then there is less updating if this changes.

bin/rad-replicate
```diff
@@ -0,0 +1,73 @@
...
+
+(def append-slash-if-missing
+  (fn [str]
+    (if (starts-with? (reverse-string str) "/")
```
MeBrei: Could be extracted to a function `ends-with?`

PR #456, opened 2019-02-18
Show diff and issue number when creating.

jameshaydon:
- The diffs and issues remote code return the new number.
- The CLI displays this new number.


jameshaydon:
Ah I didn't realise the flows had the numbers, updating.

PR #461, opened 2019-02-19
Format issue list and add show command

MeBrei:


bin/rad-issues
```diff
@@ -60,6 +61,72 @@
...
+
+(def pretty-headline
+ (fn [i]
+   (match i
+     '{:number n
+       :title tit
+       :state s
+       :username un
+       :modified-at m-a}
+     (string-append
```
jkarni: `unwords` seems slightly nicer here, but I don't feel strongly.

bin/rad-issues
```diff
@@ -60,6 +61,72 @@
...
+
+(def cells
+ {:state       10
+  :number      10
+  :title       50
+  :author      20
+  :modified-at 20})
+
+(def pad-to
+(fn [l x]
```
jkarni: Indentation

PR #463, opened 2019-02-22
macOS build + packaging

kim:
Adds a macOS build on CircleCI including a tarball deploy to Goog Clown Storage (`master` only)

Follow-up considerations:

* Will need to insert coins to keep CCI macOS executor going
* Wrapper scripts (as to-be-introduced in `f/packaging`) are not included in the tarball
* The procedure to cut a proper release is TBD. We may want to employ a different means fornaming/versioning the tarball

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -1,0 +1,124 @@
+version: 2
+jobs:
+  build:
+    macos:
+      xcode: "10.0.0"
+    steps:
+      - checkout
+
+      - restore_cache:
+          keys:
+            - stack-root-{{ arch }}-ghc8.6.3-{{ checksum "snapshot.yaml" }}-{{ checksum "package.yaml" }}
+            - stack-root-{{ arch }}-ghc8.6.3-{{ checksum "snapshot.yaml" }}
+            - stack-root-{{ arch }}-ghc8.6.3-
+
+      - restore_cache:
+          keys:
+            - toolchain-{{ arch }}
+
+      - restore_cache:
+          keys:
+            - homebrew-{{ epoch }}
+            - homebrew-
+            - homebrew
+
+      - run:
+          name: Setup Toolchain
+          command: |
+            if ! [ -x $HOME/.ghcup/bin/ghcup ]; then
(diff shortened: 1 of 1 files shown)


kim:
Some way of testing that the resulting package is non-broken may also be desirable, but could potentially be deferred to the homebrew recipe.

.circleci/config.yml
```diff
@@ -0,0 +1,126 @@
...
+      - run:
+          name: Upload to Google Cloud Storage
+          command: |
+            if [[ "$CIRCLE_BRANCH" == "master" ]]; then
+              # Key ID: db4e4a0b0ee696ea6f263c20de593c2fba4d0998
+              echo "$GCLOUD_SERVICE_KEY" | \
+                gcloud auth activate-service-account \
+                  circleci-image-uploader@opensourcecoin.iam.gserviceaccount.com \
+                  --key-file=-
+              gcloud --quiet config set project ${GCLOUD_PROJECT_ID}
```
geigerzaehler: Do we need this? If so could we hardcode the project ID or add it to the YAML file? Right now it is a secret variable which can only be discovered through the CircleCI UI and its value is hidden.

.circleci/config.yml
```diff
@@ -0,0 +1,126 @@
...
+      - run:
+          name: Package
+          command: |
+            if [[ "$CIRCLE_BRANCH" == "master" ]]; then
+              cabal v1-copy \
+                exe:rad \
+                exe:radicle \
+                exe:radicle-daemon \
+                exe:rad-machines \
+                --destdir=dist
```
geigerzaehler: How about using `dist/package-root`? I am a bit confused by putting the tarball in the package root next to `usr`.

kim: Do you mean `package-root` instead of `pkg`? The tarball is in `dist/pkg`

geigerzaehler: I mean using `--destdir=dist/package-root`, then `tar -C dist/package-root -cvzf dist/pkg/$tarball .`. Otherwise we end up with `dist/usr` and `dist/pkg` next to each other which I find confusing.

PR #469, opened 2019-02-25
Improved IPFS API exception handling

geigerzaehler:
* `ipfsHttpPost'` also uses `mapHttpException`
* Simplify pattern matching in `mapHttpException`
* Move `handleRequestException` code to `mapHttpException`
* Extract duplicated parsing code to `getJsonResponseBody`

### DIFF
--- a/src/Radicle/Ipfs.hs
+++ b/src/Radicle/Ipfs.hs
@@ -76,18 +76,24 @@ instance Exception IpfsException where
       where
         name = "Radicle IPFS daemon"
 
--- | Turn HTTP exceptions into 'IpfsException's.
+-- | Catches 'HttpException's and re-throws them as 'IpfsException's.
 mapHttpException :: forall a. IO a -> IO a
-mapHttpException io = catch io httpHdlr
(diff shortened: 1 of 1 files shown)


src/Radicle/Ipfs.hs
```diff
@@ -280,29 +285,29 @@ ipfsHttpPost
...
 
-handleRequestException :: MonadThrow m => HttpException -> m a
-handleRequestException (HttpExceptionRequest _ (ConnectionFailure _)) =
-    throw IpfsExceptionNoDaemon
-handleRequestException ex = throw ex
-
-handleParseException :: MonadCatch m => Text -> Wreq.JSONError -> m a
-handleParseException path (Wreq.JSONError msg) = throw $ IpfsExceptionInvalidResponse path (toS msg)
+-- | Parses response body as JSON and returns the parsed value. @path@
+-- if the IPFS API the response was obtained from. Throws
```
jameshaydon:
```suggestion
-- is the IPFS API the response was obtained from. Throws
```

PR #470, opened 2019-02-25
Simplify handling of pubsub message parse errors

geigerzaehler:
We simplify how parse errors for pubsub messages are logged. Before,
this was taken care of by `addHandler` and `subscribeOne`. We now
initialise the logging with `initSubscription`. This significantly
simplifies the subscription code and the logging.

### DIFF
--- a/src/Radicle/Daemon/Ipfs.hs
+++ b/src/Radicle/Daemon/Ipfs.hs
@@ -28,6 +28,7 @@ import qualified Data.Aeson.Types as Aeson
 import qualified Data.Map.Strict as Map
 import qualified Data.Sequence as Seq
 import qualified Data.Unique as Unique
+import           System.Timeout
 
 import           Radicle.Internal.Core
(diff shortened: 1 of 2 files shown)


src/Radicle/Daemon/Ipfs.hs
```diff
@@ -181,21 +187,20 @@ removeHandler :: TopicSubscription -> RegisteredHandler -> IO ()
...
-  let onMsg = \case
-        Right msg | pr msg -> putMVar var (Just msg)
-        Right _ -> pure ()
-        Left bad -> badMsg bad
-  bracket
-    (addHandler sub onMsg)
-    (removeHandler sub)
-    (const $ forkIO (threadDelay (fromIntegral timeout * 1000) >> putMVar var Nothing) *> readMVar var)
+-- | Block while waiting for a matching message.
+-- If no matching matches arrives within the given timeout (in
```
jameshaydon:
```suggestion
-- If no matching message arrives within the given timeout (in
```

PR #474, opened 2019-02-26
Package infrastructure and Debian packages

geigerzaehler:
We add the infrastructure to package Radicle and build and upload Debian
packages on CI.

* We add the `packaging/build-package.sh` script which takes care of building packages. At the moment Pacman and Debian packages can be built.
* We add `.service` files for the Radicle daemon and the Radicle IPFS daemon. These files are included in the packages.
* We change `bin/rad-daemon-ipfs` so that it can use a vendored `ipfs` binary that is included in the package.
* We build and update a Debian package on CI.


cloudbuild.yaml
```diff
@@ -206,3 +206,17 @@ steps:
...
+    - "Integration tests"
+    name: $_BUILD_IMAGE
+    env: ['STACK_ROOT=/workspace/.stack']
+    entrypoint: 'bash'
+    args:
+    - "-c"
+    - |
+      set -euxo pipefail
+
+      VERSION=g$COMMIT_SHA TARGET=debian ./packaging/build-package.sh
```
kim: What's the significance of the `g` here?

packaging/build-package.sh
```diff
@@ -0,0 +1,150 @@
...
+EOF
+
+  install-stdin "$package_root/usr/bin/radicle" <<EOF
+#!/bin/sh
+export radicle_bindir=$radicle_bindir
+export RADPATH=\${RADPATH:-$radpath}
+exec $radicle_bindir/radicle "\$@"
+EOF
+
+  install -D $(stack exec -- which git-remote-ipfs) -t "$package_root/usr/bin"
```
kim: How does this work? Is `stack` building the `git-remote-ipfs` executable implicitly?

geigerzaehler: `git-remote-ipfs` is a dependency for the tests and thus installed

https://github.com/oscoin/radicle/blob/b979ca6237185ee3b38366f63ede7c4b06c4bc88/package.yaml#L116-L117

cloudbuild.yaml
geigerzaehler: This was intended to signify a build from the **g**it repo. I have changed it to use a more explicit versioning scheme based on the date and the commit sha.

packaging/radicle-daemon.service
```diff
@@ -0,0 +1,11 @@
+[Unit]
```
xla: Should we not express that this service `Wants` the radicle-ipfs sefvice?

geigerzaehler: We define `Requires=radicle-ipfs.service` below. This is a stronger version of [`Wants=`](https://www.freedesktop.org/software/systemd/man/systemd.unit.html#Wants=)

README.md
```diff
@@ -82,6 +82,17 @@ with the output of `docker-machine ip`.
 You can reset the test daemon’s machine configuration by removing the file
 `/tmp/radicle-machines.json`.
 
+### Packaging
+
+Packages can be build with the `./packaging/build-package.sh` script. Run it
```
jameshaydon:
```suggestion
Packages can be built with the `./packaging/build-package.sh` script. Run it
```

README.md
```diff
@@ -82,6 +82,17 @@ with the output of `docker-machine ip`.
...
+with `-h` for more information. The script requires [`fpm`][fpm].
+
+On CI a Debian package is build for every commit and uploaded to
```
jameshaydon:
```suggestion
On CI a Debian package is built for every commit and uploaded to
```

PR #476, opened 2019-02-26
Improve help texts of the diff, issue & create-keys app

MeBrei:
More detailed description of the diff/issue/create-keys app commands. Also in regard to the idea of using the help texts in the documentation.

### DIFF
--- a/bin/rad-issue
+++ b/bin/rad-issue
@@ -22,22 +22,31 @@
 
 (def help
   (string-append
-  "rad-issue - Radicle issue CLI
+  "rad issue - Radicle Issue CLI
 
    Usage:
-        rad-issue list " list-opts-help "
-        rad-issue new
-        rad-issue [show | close] <issue-number>
-        rad-issue comment <issue-number> <comment>
-        rad-issue help
+        rad issue list " list-opts-help "
+        rad issue new
+        rad issue [show | close] <issue-number>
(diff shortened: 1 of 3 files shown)


bin/rad-issue
```diff
@@ -22,22 +22,31 @@
...
-
-     list         - Select from issues to view or comment on
+        rad issue list " list-opts-help "
+        rad issue new
+        rad issue [show | close] <issue-number>
+        rad issue comment <issue-number> <comment>
+        rad issue help
+
+     list         - Lists all issues
+                    The list can be filtered by state:
```
jameshaydon: Maybe extract this string to a function to avoid duplication with the rad-diff help.

MeBrei: I would leave it for now. Maybe we could have a future PR where we also extract all non-error user feedback to make them more consistent.

xla:
In the user flows we had `rad key create` to leave the possibilty to extend the namespace further down the road (listing and general management). Is there strong reason to not move to that?

MeBrei:
> In the user flows we had `rad key create` to leave the possibilty to extend the namespace further down the road (listing and general management). Is there strong reason to not move to that?

@xla No strong reason against that. I will create a ticket for it and change it in a different PR as there will be more changes than only the help text.

PR #477, opened 2019-02-26
Align macOS packaging

geigerzaehler:
The macOS package build now uses `./packaging/build-package.sh`. This changes the file layout in the tarball significantly.

* We switch from `cabal build` to `stack build` to align with the other package builds.
* We include `git-remote-ipfs` in the package. This requires us to build the test dependencies.
* We create an upload a package for every build.
* We rename the macOS tarball to the same format as the other packages. I.e. `radicle_YYYY-MM-DD-<commitsha>_x86_64-darwin`

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -54,11 +54,8 @@ jobs:
           command: |
             brew install coreutils
             echo 'export PATH="/usr/local/opt/coreutils/libexec/gnubin:$PATH"' >> $BASH_ENV
-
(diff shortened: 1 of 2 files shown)


geigerzaehler:
> Obviously, I'm not a big fan of leaving a broken `Paths_radicle` in place and juggling environment variables instead 🤷‍

Yeah, it’s not ideal. I hope we can move to a cabal-based build in the future so that we can take advantage of that configuration.

> One thing you might consider is to copy the LICENSE file into the packages (cabal copy does this), as it says one is obliged to do so.

Will do

PR #480, opened 2019-02-27
Extract error messages for command parsing

MeBrei:
This is the first step of extracting all user facing error messages. Once we agree on the format, this should be extended to the apps and other parts that produce user facing errors.

My idea here was that all user facing errors (produced by radicle code) are collected here and separated into section (not by file where they are used, but by topic). Each function for an error message has a simple test. Like this, changes can be tested locally with `stack exec -- radicle test/all.rad`, both the tests from `error-messages.rad` and where they are actually used.

### DIFF
--- a/rad/prelude.rad
+++ b/rad/prelude.rad
@@ -15,7 +15,6 @@
 (file-module! "prelude/ref.rad")
 (file-module! "prelude/lens.rad")
 (file-module! "prelude/key-management.rad")
-(file-module! "prelude/error-messages.rad")
 (file-module! "prelude/machine.rad")
 (file-module! "prelude/machine-remote.rad")
 (file-module! "prelude/state-machine.rad")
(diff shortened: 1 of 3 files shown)


MeBrei:
Moved `error-messages.rad` out of prelude and added some descriptions, how to run tests for the file and "conventions" on how to write the functions.

xla:
@MeBrei This looks great!  Could we add a doc to the error functions which gives some context when and how the error(s) are used?

MeBrei:
> Could we add a doc to the error functions which gives some context when and how the error(s) are used?

@xla In the future, at least for functions where the name is not descriptive enough I would add a doc. The functions here seemed to me simple enough to work without doc.

xla:
> @xla In the future, at least for functions where the name is not descriptive enough I would add a doc. The functions here seemed to me simple enough to work without doc.
> The documentation can be extended in the future as well.

That might be true for someone familiar with the code and its use. My perspective was from someone checking/adjusting error strings from a UX perspective and without more contexxt will be hard pressed to anticipate in what scenarios an error might occur.

PR #481, opened 2019-02-27
Set STACK_ROOT when building daemon image

geigerzaehler:
We set the `STACK_ROOT` environment variable for the “Build rad-daemon-radicle” build step. The build steps uses `stack` and without setting `STACK_ROOT` it will install GHC.


geigerzaehler:
This reduces the build time by roughly 90 seconds.

PR #487, opened 2019-02-27
Extract logging into separate module

geigerzaehler:
Extract all logging code into `Radicle.Daemon.Logging`.

This also gets rid of the extra `LogLevel` argument to `logInfo`. `logInfo Normal` becomes `logInfo`, `logInfo Debug` becomes `logDebug`.


geigerzaehler:
> Still a little unclear to me why we don't just use a library, bu

We should at some point. But to be able to use a library we first need clear separation between the logging logic and the rest of the application

PR #492, opened 2019-02-28
daemon: Wait for IPFS daemon

geigerzaehler:
With systemd the Radicle daemon is started very quickly after the IPFS daemon has started. Therefore the IPFS daemon might not be ready yet when the Radicle daemon tries to load the machines. This results in errors.

To prevent this the Radicle daemon waits until the IPFS daemon is ready by repeatedly trying to connect to its `version` endpoint.

### DIFF
--- a/exe/Daemon.hs
+++ b/exe/Daemon.hs
@@ -7,8 +7,9 @@
 -- the RFC>.
 module Daemon (main) where
 
-import           Protolude hiding (fromStrict, option, poll)
+import           Protolude hiding (catch, fromStrict, option, poll, tryJust)
 
+import           Control.Exception.Safe hiding (Handler)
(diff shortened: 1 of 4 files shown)


kim:
I believe the most common way to implement readiness probes with systemd (in lieu of a socket-activated service, or one that properly `fork`s when  ready) is to specify `ExecStartPost`. A `curl --connect-timeout ...` should do

geigerzaehler:
> I believe the most common way to implement readiness probes with systemd (in lieu of a socket-activated service, or one that properly `fork`s when ready) is to specify `ExecStartPost`. A `curl --connect-timeout ...` should do

Nice! I’ll add that to the systemd file

PR #500, opened 2019-03-01
Extract errors from diff and issue app

MeBrei:


rad/prelude/error-messages.rad
```diff
@@ -260,3 +260,16 @@ This may be because you have no internet connection, you are not authorized to p
...
+;; key management
+
+(def missing-key-file
+  "Any request to the machine is aborted, when the key file can't be found."
+  (fn []
+    (string-append
+      "Missing file: my-keys.rad\n"
+      "Do `rad create-keys` to create your own key pair and rerun your command.")))
+
+(:test "missing-key-file"
```
geigerzaehler: I don’t think we need the tests. Their quite trivial

MeBrei: I agree that they are trivial. The idea is if people change the text (maybe even adding arguments to the function) then there is a quick way of testing that the function is still working and giving the expected result. Especially if someone not so familiar with radicle is editing the texts.

PR #503, opened 2019-03-01
Rename rad-create-keys to rad-key

MeBrei:


bin/rad-key
```diff
@@ -5,13 +5,13 @@
...
+  "rad key - Radicle Key Management CLI
 
    Usage:
-        rad create-keys [--force]
-        rad create-keys help
+        rad key create [--force]
+        rad key help
 
-   Create a new key pair with `rad create-keys`. If `--force` is appended, an
+   Create a new key pair with `rad key create`. If `--force` is appended, an
```
geigerzaehler: Giving the command `rad key create` again seems redundant.

PR #505, opened 2019-03-01
Always exit with error if key is missing

geigerzaehler:
We make sure that all commands show an error message and exit if the user keys are missing. Before we would throw a cryptic exception because `read-keys!` returned `:nothing`.

### DIFF
--- a/rad/prelude/machine.rad
+++ b/rad/prelude/machine.rad
@@ -12,10 +12,7 @@
 (import prelude/patterns :unqualified)
 (import prelude/strings :unqualified)
 (import prelude/seq :unqualified)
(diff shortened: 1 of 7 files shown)


rad/prelude/machine.rad
```diff
@@ -12,10 +12,7 @@
 (import prelude/patterns :unqualified)
 (import prelude/strings :unqualified)
 (import prelude/seq :unqualified)
-(import prelude/key-management :as 'key-mgmt)
-
-(file-module! "prelude/error-messages.rad")
-(import prelude/error-messages :as 'error)
+(import prelude/key-management '[get-keys])
```
MeBrei: `get-keys` is missing the `!` here

PR #510, opened 2019-03-04
Change radicle_bindir for darwin.

jameshaydon:


packaging/build-package.sh
```diff
@@ -130,6 +130,7 @@ function package-debian () {
 }
 
 function package-darwin () {
+  radpath=/usr/local/lib/radicle/modules
```
jkarni: Should we move all the radpaths into their respective functions? Otherwise, changing the order of invocations slightly will break things.

PR #514, opened 2019-03-05
Make pattern matching work in prelude modules

geigerzaehler:
This fixes a bug where one of the prelude modules uses extended pattern matching (as defined in `match-pat`) without importing `prelude/patterns`.

For example one of the prelude modules may contain the code

    (def foo
      (match bar
        :foo #t
        _    #f))

If `match-pat` from `prelude/patterns` is not in the global environment when this module is evaluated then the `match` special form will call the matcher `:foo` as a function which will throw an error.

### DIFF
--- a/rad/prelude.rad
+++ b/rad/prelude.rad
@@ -3,6 +3,10 @@
 
 (file-module! "prelude/basic.rad")
 (file-module! "prelude/patterns.rad")
+;; Adds the special function `match-pat` to the global scope. This is
+;; required for the extended `match` to work in the modules loaded
+;; after.
+(import prelude/patterns :unqualified)
 (file-module! "prelude/bool.rad")
 (file-module! "prelude/seq.rad")
 (file-module! "prelude/strings.rad")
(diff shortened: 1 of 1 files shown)


jameshaydon:
The prelude modules which require pattern matching usually require it. E.g.:

```
{:module 'prelude/strings
 :doc    "String manipulation functions."
 :exports '[intercalate unlines unwords split-by words lines map-string
            reverse-string ends-with?] }

(import prelude/basic :unqualified)
(import prelude/seq :unqualified)
(import prelude/patterns :unqualified)

...
```

Will this work for the module in question? I think it might be best if the prelude submodules make dependencies explicit.

geigerzaehler:
> Will this work for the module in question? I think it might be best if the prelude submodules make dependencies explicit.

Adding `(import prelude/patterns :unqualified)` would indeed fix the problem. But having the import as an implicit dependency for using `match` is not great. Especially since the error message does not help at all. I spend an hour debugging this.

PR #515, opened 2019-03-05
rad-project: Error when 'rad ipfs' fails

geigerzaehler:
We show an error and exit when the `rad ipfs` commands used by `rad project` fail to work.

### DIFF
--- a/bin/rad-project
+++ b/bin/rad-project
@@ -39,15 +39,22 @@
   (fn [orig-cid]
     (def keyname (uuid!))
     (def key
-      (unlines
-        (process-with-stdout!
-          "rad"
-          ["ipfs" "key" "gen" "--type=ed25519" keyname]
-          "")))
-    (process-with-stdout!
-      "rad"
-      ["ipfs" "name" "publish" "--key" key orig-cid]
-      "")
+      (match
+        (process-with-stdout-stderr-exitcode!
+          "rad" ["ipfs" "key" "gen" "--type=ed25519" keyname] "")
+        ['stdout _ :ok] (unlines stdout)
+        ['stdout 'stderr _]
+          (do
+            (put-str! (error/rad-ipfs-key-gen-failure (unlines stderr)))
+            (exit! 1))))
+    (match
+      (process-with-stdout-stderr-exitcode!
+        "rad" ["ipfs" "name" "publish" "--key" key orig-cid] "")
+      ['stdout _ :ok] :ok
+      ['stdout 'stderr _]
+        (do
+          (put-str! (error/rad-ipfs-name-publish-failure (unlines stderr)))
+          (exit! 1)))
     (def remote (string-append "ipfs://ipns/" key))
     remote))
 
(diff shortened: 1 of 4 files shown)


bin/rad-project
```diff
@@ -39,15 +39,24 @@
...
+            (exit! 1))))
+    (match
+      (process-with-stdout-stderr-exitcode!
+        "rad" ["ipfs" "name" "publish" "--key" key orig-cid] "")
+      ['stdout _ :ok] :ok
+      ['stdout 'stderr _]
+        (do
+          (put-str! "error: Failed to publish key for Git IPFS repository:")
+          (put-str! (unlines stderr))
+          (exit! 1)))
```
MeBrei: both user facing errors should be collected in in `error-messages.rad`

rad/prelude/io.rad
```diff
@@ -116,7 +116,8 @@ Example: `(shell-no-stdin! \"ls -Glah\")`. "
 
 (def process!
   "Executes `command` using `execvp` with `to-write` as input. Stdout and stderr are inherited.
-See `man exec` for more information on `execvp`.
+See `man exec` for more information on `execvp`. Returns `:ok` if the
+process exited normaly and `[:error n]` otherwise.
```
MeBrei: small typo: should be `normally`

rad/prelude/error-messages.rad
```diff
@@ -273,3 +275,15 @@ This may be because you have no internet connection, you are not authorized to p
...
+  (fn [stderr]
+    (string-append
+      "error: Failed to create key for Git IPFS repository:\n"
+      stderr)))
+
+(def rad-ipfs-name-publish-failure
+  (fn [stderr]
+    (string-append
+      "error: Failed to publish key for Git IPFS repository:\n"
+      stderr)))
```
MeBrei: Please add tests and a documentation on where the errors occur and when.

geigerzaehler: I added documentation but I’m not a big fan of adding tests here. The logic is so trivial that the tests are basically copy-and-pasting the function body. This does not prevent errors and just adds maintenance overhead.

MeBrei: If someone not so familiar with radicle is supposed to update the texts, I think it would be good for them to have a quick way of verifying that the functions are still working and give the proper results.

MeBrei: Can you think of a better way for this scenario?

geigerzaehler: Ah, you’re right. I didn’t consider people not familiar with Radicle. I added the tests

rad/prelude/error-messages.rad
```diff
@@ -273,3 +275,26 @@ This may be because you have no internet connection, you are not authorized to p
...
+(def rad-ipfs-key-gen-failure
+  "Printed when the `rad ipfs key gen` command in `init-git-ipfs-repo`
+  in `rad-project` fails. Takes stderr of the command as an argument."
+  (fn [stderr]
+    (string-append
+      "error: Failed to create key for Git IPFS repository:\n"
+      stderr)))
+
+(:test "rad-ipfs-key-gen-failure"
+  [("rad-ipfs-key-gen-failure" "stderr") ==>  "error: Failed to create key for Git IPFS repository:\nstderr"])
```
MeBrei: `"` around the function name need to be removed.

PR #518, opened 2019-03-06
Extract pad-to from issue/diff app

MeBrei:


rad/prelude/strings.rad
```diff
@@ -107,6 +108,28 @@
   [ (lines " \n \t") ==> [" " " \t"]]
   [ (lines "") ==> []])
 
+;; pad-to
+
+(def pad-to
+  "Appends the `word` with whitespace to get to length `l`. If `word` is longer
```
jameshaydon: Don't most "padding" function just return the input rather than truncate?

MeBrei: You are right. We need the truncate-or-pad functionality to format the list tables. Would `pad-or-truncate-to` work, or can you think of a shorter function name?

jameshaydon: I would probably implement `pad` and then just use `(take _ (pad _))` at the call site. If it is used many times maybe call it `pad-truncate`.

MeBrei: I like the `(take _ (pad _))` option. It is used only twice so far. A more accurate naming would be `pad-right-to` probably.

PR #521, opened 2019-03-06
Use MonadUnliftIO for Daemon

geigerzaehler:
`Daemon` gets a `MonadUnliftIO` instance and we generalize the functions in `Radicle.Internal.ConcurrentMap` to use the `MonadUnliftIO` constraint instead of `IO`.

The immediate benefit is that `Daemon.modifyMachine` is simplified: The control logic is extracted and we focus on the domain logic.

The change will also make similar patterns easier to implement in the future.

### DIFF
--- a/exe/Daemon.hs
+++ b/exe/Daemon.hs
@@ -369,20 +369,19 @@ insertNewMachine id m = do
 -- machine isn't in the cache already.
 modifyMachine :: forall a. MachineId -> (Machine -> Daemon (Machine, a)) -> Daemon a
 modifyMachine id f = do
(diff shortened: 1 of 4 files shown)


exe/Daemon.hs
```diff
@@ -370,19 +370,18 @@ insertNewMachine id m = do
 modifyMachine :: forall a. MachineId -> (Machine -> Daemon (Machine, a)) -> Daemon a
 modifyMachine id f = do
     env <- ask
```
jameshaydon: This can probably be `asks machines` now.

PR #523, opened 2019-03-07
Add more options for rad version command

MeBrei:
Adds `rad --version` & `rad -v` (returns the same as `rad version`)

### DIFF
--- a/exe/Rad.hs
+++ b/exe/Rad.hs
@@ -28,7 +28,11 @@ runCommand [] = do
     putStrLn @Text "Available commands:"
     putStr $ T.unlines $ ["   " <> T.pack c | c <- availableCommands]
 runCommand ("version":_) =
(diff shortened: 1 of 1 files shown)


exe/Rad.hs
```diff
@@ -28,7 +28,11 @@ runCommand [] = do
     putStrLn @Text "Available commands:"
     putStr $ T.unlines $ ["   " <> T.pack c | c <- availableCommands]
 runCommand ("version":_) =
-    putStrLn ("radicle version " <> toS (showVersion version) :: Text)
+    putStrLn radicleVersion
+runCommand ("--version":_) =
+    putStrLn radicleVersion
+runCommand ("-v":_) =
+    putStrLn radicleVersion
```
geigerzaehler: You could extract the `putStrLn` call, too.

PR #525, opened 2019-03-07
rad-project: Add other git remote

geigerzaehler:
If the user wants to add their own remote on `project init` we actually add the input URL as a git remote. We also provide tests for this.

### DIFF
--- a/rad/prelude/error-messages.rad
+++ b/rad/prelude/error-messages.rad
@@ -11,6 +11,7 @@
    parent-commit-not-master checkout-new-branch-failure checkout-master-failure
    applying-patch-failure applying-accepted-patch-failure push-patch-failure
    missing-key-file rad-ipfs-name-publish-failure rad-ipfs-key-gen-failure
+   process-exit-error
    ]}
 
 (import prelude/strings :unqualified)
@@ -18,6 +19,7 @@
 ;; parsing
 
 (def missing-arg
+  "Used for command line parsing when an argument to a command is missing."
   (fn [arg cmd]
     (string-append "Missing argument \"" arg "\" for command \"" cmd "\"")))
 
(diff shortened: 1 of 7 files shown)


rad/prelude/io.rad
```diff
@@ -189,6 +192,23 @@ process exited normally and `[:error n]` otherwise. Example: `(process!
...
+(def process-with-stdout-strict!
+  "Like `process-with-stdout!`, but prints an error message and exits if the command fails."
+  (fn [command args to-write]
+    (match (process-with-stdout-stderr-exitcode! command args to-write)
+      ['stdout _ :ok] stdout
+      ['stdout 'stderr [:error 'n]]
+        (do
+          (def command-line (strings/intercalate " " (cons command args)))
+          (put-str! (string-append "error: Command \"" command-line "\" exited with status code " (show n)))
+          (map (fn [l] (put-str! (string-append "  stderr: " l))) stderr)
```
MeBrei: As this is a user facing error, it should be moved to `error-messages.rad`. It might also make sense to pass part of the error message to the function so that we can print specific instructions for the user, but it depends on where it is/will be used.

geigerzaehler: I extracted the error message but I’m not sure about the second part. Currently the function is only used in one place where it does not make sense to pass an error message. If we need that functionality in the future we can add it.

rad/prelude/error-messages.rad
```diff
@@ -309,3 +310,20 @@ Do `rad key create` to create your own key pair and rerun your command."])
...
+    (def message (string-append "error: Command \"" command-line "\" exited with status code " (show exit-code)))
+    (def stderr-output (map (fn [l] (string-append "  stderr: " l)) stderr))
+    (unlines (cons message stderr-output))
+    ))
+
+(:test "process-exit-error"
+  [(process-exit-error "foo" ["arg1" "arg2"] 5 ["fail1" "fail2"]) ==>
+"error: Command \"foo arg1 arg2\" exited with status code 5
+  stderr: fail1
+  stderr: fail2"])
```
MeBrei: For better organization, add a comment as a header for this messages. Currently it is placed under `;; key management`

MeBrei:
👍

PR #527, opened 2019-03-07
Better concurrency properties for ConcurrentMap

geigerzaehler:
We improve contention of `modifyValue`. Before, two actions `modifyValue k1 f1` and `modifyValue k2 f2` would always contend. Now they can run
concurrently if `k1` and `k2` are different. Only if `k1 == k2` does one action wait for the other.

The change also allows `modifyExistingValue` to use `modifyValue` without loosing any of its concurrency properties.

While `modifyValue` is not currently used this change allows us to use it instead of `insert` in the future. This will prevent race conditions where multiple threads insert a value in the map.

### DIFF
--- a/src/Radicle/Internal/ConcurrentMap.hs
+++ b/src/Radicle/Internal/ConcurrentMap.hs
@@ -1,6 +1,5 @@
 module Radicle.Internal.ConcurrentMap
   ( empty
-  , fromMap
   , lookup
   , nonAtomicRead
   , insert
@@ -14,26 +13,23 @@ import           Protolude hiding (empty)
 import           Control.Monad.IO.Unlift
 import qualified Data.Map.Strict as Map
 
--- | A Map which offers atomic operations on the values associated to
--- keys. Assumes that the values are independent of one another; does
--- not offer a consistent view over the values. Therefore this is best
--- used when the keys uniquely identify the resources represented by
--- the values.
-newtype CMap k v = CMap (MVar (Map k (MVar v)))
+-- | A mutable Map that serializes operations on the same key.
+--
+-- Calls to the 'lookup', 'insert', 'modifyValue', and
+-- 'modifyExistingValue' with the same key are serialized. Calls to
+-- these functions with different keys can run concurrently.
+newtype CMap k v = CMap (MVar (Map k (MVar (Maybe v))))
 
 -- | Create a new empty 'CMap'.
 empty :: IO (CMap k v)
 empty = CMap <$> newMVar Map.empty
-
-fromMap :: Map k v -> IO (CMap k v)
-fromMap m = CMap <$> (traverse newMVar m >>= newMVar)
 
 -- | Atomically lookup a value.
 lookup :: Ord k => k -> CMap k v -> IO (Maybe v)
 lookup k (CMap m_) = withMVar m_ $ \m -> do
    case Map.lookup k m of
      Nothing -> pure Nothing
-     Just v_ -> pure <$> readMVar v_
+     Just v_ -> readMVar v_
 
 -- | Non-atmoically read the contents of a 'CMap'. Provides a
 -- consistent shapshot of which keys were present at some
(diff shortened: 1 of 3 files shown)


src/Radicle/Internal/ConcurrentMap.hs
```diff
@@ -14,67 +13,71 @@ import           Protolude hiding (empty)
...
+    -- Only one thread can take the value of a key. All other calls to
+    -- 'takeValue' block until the 'MVar' has been filled.
+    takeValue :: k -> IO (MVar (Maybe v), Maybe v)
+    takeValue k' = modifyMVar m_ $ \m ->
+        case Map.lookup k' m of
+            Nothing -> do
+                valueVar <- newEmptyMVar
+                pure (Map.insert k' valueVar m, (valueVar, Nothing))
+            Just valueVar -> do
+                maybeV <- readMVar valueVar
```
jameshaydon: Should this be `takeMVar`?

src/Radicle/Internal/ConcurrentMap.hs
```diff
@@ -14,67 +13,71 @@ import           Protolude hiding (empty)
...
+-- | Atomically modifies a value associated to a key. Blocks all
+-- operations on @k@ while @f@ is running.
+modifyValue :: forall k v a m. (MonadUnliftIO m, Ord k) => k -> CMap k v -> (Maybe v -> m (Maybe v, a)) -> m a
+modifyValue k (CMap m_) f = withRunInIO $ \runInIO -> mask $ \restore -> do
+    (valueVar, maybeValue) <- takeValue k
+    (maybeValue', a) <- restore (runInIO $ f maybeValue) `onException` putMVar valueVar maybeValue
+    -- To prevent space leaks we probably should remove the key from
+    -- the map if @maybeValue'@ is 'Nothing'. I’m not sure whether this
+    -- might lead to incorrect behavior, though.
+    putMVar valueVar maybeValue'
```
jameshaydon: I think we can delete in case of `Nothing` after this put? What makes you unsure?

src/Radicle/Internal/ConcurrentMap.hs
```diff
@@ -14,67 +13,71 @@ import           Protolude hiding (empty)
 import           Control.Monad.IO.Unlift
 import qualified Data.Map.Strict as Map
 
--- | A Map which offers atomic operations on the values associated to
--- keys. Assumes that the values are independent of one another; does
--- not offer a consistent view over the values. Therefore this is best
--- used when the keys uniquely identify the resources represented by
--- the values.
-newtype CMap k v = CMap (MVar (Map k (MVar v)))
+-- | A mutable Hash-Map that serializes operations on the same key.
```
jameshaydon: Isn't it just a mutable map rather than a hash-map?

src/Radicle/Internal/ConcurrentMap.hs
```diff
@@ -14,67 +13,61 @@ import           Protolude hiding (empty)
...
+modifyValue :: forall k v a m. (MonadUnliftIO m, Ord k) => k -> CMap k v -> (Maybe v -> m (Maybe v, a)) -> m a
+modifyValue k (CMap m_) f = withRunInIO $ \runInIO -> do
+    valueVar <- modifyMVar m_ $ \m ->
+        case Map.lookup k m of
+            Nothing -> do
+                valueVar <- newMVar Nothing
+                pure (Map.insert k valueVar m, valueVar)
+            Just valueVar -> do
+                pure (m, valueVar)
+    -- To prevent space leaks we probably should remove the key from
```
jameshaydon: With this new version I think the only way to get rid of the `Nothing`s would be to schedule a cleanup. We don't have a "delete machine" endpoint so it's really not a problem for us (yet!).

test/spec/Radicle/Internal/ConcurrentMapTest.hs
```diff
@@ -0,0 +1,46 @@
...
+
+import           Control.Concurrent.Async
+import           Protolude
+
+import qualified Radicle.Internal.ConcurrentMap as CMap
+import           Test.Tasty
+import           Test.Tasty.HUnit
+
+test_ :: TestTree
+test_ = testGroup "Radicle.Internal.ConcurrentMap"
```
jameshaydon: It would be nice to have a test for a `CMap Bool Int` with some `threadDelay`s which show (because of total time elapsed) that there is no contention if the key is `/=`, but I'm fine to merge as is.

geigerzaehler: I have added a test for this. Could you have another look?

PR #529, opened 2019-03-07
Verify patch is on top of origin/master

MeBrei:


bin/rad-diff
```diff
@@ -99,6 +99,14 @@
   (fn [commit]
     (string/unlines (process-with-stdout! "git" ["rev-parse" "--short" commit] ""))))
 
+(def get-commit-parent!
+  (fn [commit]
+    (string/unlines (process-with-stdout! "git" ["rev-parse" (string-append commit "^")] ""))))
+
+(def get-head-origin-master!
+  (fn []
+    (string/unlines (process-with-stdout! "git" ["rev-parse" "origin/master"] ""))))
```
geigerzaehler: Maybe you could extract 

~~~clojure
(def git-rev-parse
  (fn [commitish] (string/unlines (process-with-stdout! "git" ["rev-parse" commitish] "")))
~~~

And then inline the calls to this function in `verify-patch`.

rad/prelude/error-messages.rad
```diff
@@ -144,6 +144,17 @@ Set the upstream branch manually and rerun 'rad project init'"])
 (:test "unknown-commit"
   [(unknown-commit "12345") ==> "Commit 12345 can't be found."])
 
+(def invalid-patch
```
geigerzaehler: We could be a bit more explicit in the name. Maybe `parent-commit-not-master`?

bin/rad-diff
MeBrei: I think this should be done in a different PR as there are more git rev parse commands (with options) and git commands in general could be more cleaned up.

PR #542, opened 2019-03-12
Item newness module

jameshaydon:
For the moment this just tracks opening of issues and patches, and displays a `*` to mark unread items in lists.

A subsequent PR will add filtering by new.


rad/prelude/track-newness.rad
```diff
@@ -0,0 +1,40 @@
...
+           (/just 'x) [:read x])))
+
+(def read-since-modified!?
+  "Returns true if an item has not been read, or if it has been modified since it
+  was last read. Assumes the item has a `:number` field for a unique ID, and a
+  `:modified-at` field containing the datetime at which it was last modified."
+  (fn [machine-id item]
+    (match (get-item-newness! machine-id (lookup :number item))
+           :unread #f
+           [:read 't] (<= (unix-epoch (lookup :modified-at item)) (unix-epoch t)))))
```
MeBrei: This will not catch added comments. Currently the items `modified-at` doesn't change if a comment is added.

jameshaydon: Good point! But this seems like an issue with maintaining `:modified-at` correctly rather than the current PR?

jameshaydon: `Created issue #5 in 12D3KooWP7mz4WKrAwN9LXymnwntxMxj7sUYMCaWodX3EUDWmuVD`

PR #551, opened 2019-03-13
Don’t load failed machines over and over

geigerzaehler:
Before we tried to reload machines that failed to initialize every ten seconds. This included invalid machines that failed to evaluate properly. We disable this behavior until we have a better strategy to catch errors and have exponential back-off.


geigerzaehler:
> We should also distinguish between machines that have failed to initialise because of IPNS lookup failure, or another error.

Yes. For the former we should retry.

PR #552, opened 2019-03-13
Check inputs are valid before IPFS write.

jameshaydon:


exe/Daemon.hs
```diff
@@ -438,8 +438,8 @@ installMachineMessageHandler m handleMessage = do
 writeInputs :: MachineId -> [Value] -> Maybe Text -> Daemon [Value]
 writeInputs id inputs nonce = do
     (rs, idx) <- modifyMachine id $ \machine -> do
-        newIndex <- machineIpfs id $ writeIpfs id inputs
         (machine', results) <- addInputs inputs newIndex machine
+        newIndex <- machineIpfs id $ writeIpfs id inputs
```
geigerzaehler: This does not work because we need to known the new index. Maybe it would make sense to split `writeIpfs` in two parts: One that just writes the new inputs to IPFS but does not update the IPNS record. The second part updates the IPNS record.

jameshaydon: Why write bad inputs to IPFS? Can we go back to the implementation before https://github.com/radicle-dev/radicle/commit/24db69a36fff7b58bd4425edb1b6824af42ff8a3?

geigerzaehler: That would work, too

jameshaydon:
@geigerzaehler Can you check again.

exe/Daemon.hs
```diff
@@ -305,30 +305,35 @@ loadMachine mode id = do
...
-  -- ^ Inputs to add.
-  -> MachineEntryIndex
-  -- ^ The new index.
-  -> Machine
-  -> Daemon (Machine, [Value])
-  -- ^ Returns the updated machine and the produced outputs
-addInputs inputs index m = do
+-- | Checks that inputs can be applied to a machine. Returns the new state that
+-- would result, and the outputs.
+checkInputs :: Machine -> [Value] -> Daemon (Bindings (PrimFns Identity), [Value])
```
geigerzaehler: How about calling this `runInputs`? And update the docs to mention that it throws when an error when evaluation fails.

PR #557, opened 2019-03-13
Explicit timeouts for IPFS API.

jameshaydon:


geigerzaehler:
@jameshaydon Could we delay this until https://github.com/radicle-dev/radicle/pull/571 is merged? We will need to re-implement it after the switch from `wreq` to `http-client`

PR #560, opened 2019-03-14
fixed Radicle

0xflotus:


geigerzaehler:
Thanks @0xflotus. Could you rebase your change? That should also fix the build.

PR #564, opened 2019-03-14
Patterns for parsing whole numbers and simpler cmd parsing

jameshaydon:
And use them in command parsing to avoid the repeated nested `if`s in all the
functions.

### DIFF
--- a/bin/rad-patch
+++ b/bin/rad-patch
@@ -362,6 +362,10 @@
 
 (def patch-num-pat (/number-string (/integer 'patch-number)))
 
+(def patch-commands (/member ["checkout" "show" "retract" "accept" "reject" "comment"]))
+
+(/cmd "checkout" args)
+
 (machine/catch-daemon!
  (fn []
    (def bad-num (fn [] (cmd-parse-failure (error/whole-item-number :patch))))
--- a/bin/rad-patch
+++ b/bin/rad-patch
@@ -316,52 +316,37 @@
 (def show-patch!
   "Shows a single PATCH `n`"
   (fn [machine n]
-    (if (number? n)
-      (if (integral? n)
-        (do
-          (def patch (verify-patch-number n machine))
-          (put-str! (pretty-patch patch)))
-        (cmd-parse-failure (error/whole-item-number :patch)))
-      (cmd-parse-failure (error/missing-item-number :patch "show")))))
+    (def patch (verify-patch-number n machine))
+    (put-str! (pretty-patch patch))))
 
 (def edit-patch-state!
   "Edit the state of a PATCH. If the new state is `:accepted` the PATCH is then
   also merged to master and the changes pushed."
   (fn [machine n new-state]
-    (if (number? n)
-      (if (integral? n)
-        (do
-          (def edit-state-with-feedback!
-            (fn [message]
-              (catch 'daemon-error
-                (simple-edit-patch! machine n {:state new-state})
-                (fn [x]
-                   (put-str!
-                     (error/state-change-failure :patch (pretty-state new-state)))
-                   (exit! 1)))
-              (put-str! message)))
-          (verify-patch-number n machine)
-          (if (eq? new-state :accepted)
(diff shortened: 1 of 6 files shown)


rad/prelude/patterns.rad
```diff
@@ -227,3 +228,29 @@
...
+      (if (integral? v)
+        (match-pat pat v)
+        :nothing))))
+
+(def /number-string
+  "Matches strings which can be parsed as numbers. The parsed number is passed to
+  the inner pattern."
+  (fn [pat]
+    (fn [s]
+      (def n (read s))
```
jkarni: Maybe a `catch` here, since otherwise we get a nothing when it parses as something other than number, but an exception if it doesn't parse as anything at all, which seems confusing.

rad/prelude/patterns.rad
```diff
@@ -227,3 +228,29 @@
           [1 (/member [1 3 5])] :no
           [1 (/member [2 4 8])] :yes
           _                     :no) ==> :yes])
+
+(def /number
+  "Matches values which are numbers. Matches are passed to the inner pattern."
+  (fn [pat]
+    (fn [v]
+      (if (number? v)
+        (match-pat pat v)
```
jkarni: Interesting pattern (this passing into inner patterns). I see the point of it for the usage above, and am happy with it, but I'm also curious about what the more general principle is (e.g., if the *pattern* is just the number 7, for example, it shouldn't bind an x; intuitively that might mean "any predicate that has more than a one-element extension", but I'm not sure)

rad/prelude/patterns.rad
```diff
@@ -227,3 +228,31 @@
...
+(def /number-string
+  "Matches strings which can be parsed as numbers. The parsed number is passed to
+  the inner pattern."
+  (fn [pat]
+    (fn [s]
+      (catch 'any
+          (do (def n (read s))
+              (if (number? n)
+                (match-pat pat n)
+                :nothing))
```
MeBrei: You could reuse `/number` here.

rad/prelude/patterns.rad
jameshaydon: If the inner pat is `7`, that is `(/number 7)`, then it matches `7`. Any other number would match the outer pattern but then fail at the pattern `7`.

I'm not sure what you mean by "bind an `x`" -- there is no `x` here.

bin/rad-patch
```diff
@@ -375,17 +360,25 @@
...
+          (/cmd-1 "show" patch-num-pat help) (show-patch! (machine-name) patch-number)
+          (/cmd-1 "show" _ help)                (cmd-parse-failure (error/whole-item-number :patch))
+          (/cmd-1 "retract" patch-num-pat help) (edit-patch-state! (machine-name) patch-number :retracted)
+          (/cmd-1 "retract" patch-num-pat help) (cmd-parse-failure (error/whole-item-number :patch))
+          (/cmd-1 "accept" patch-num-pat help) (edit-patch-state! (machine-name) patch-number :accepted)
+          (/cmd-1 "accept" _ help)                (cmd-parse-failure (error/whole-item-number :patch))
+          (/cmd-1 "reject" patch-num-pat help) (edit-patch-state! (machine-name) patch-number :rejected)
+          (/cmd-1 "reject" _ help)                (cmd-parse-failure (error/whole-item-number :patch))
+          (/cmd-2 "comment" patch-num-pat 'comment help) (create-comment! (machine-name) patch-number comment)
+          (/cmd-2 "comment" _ 'comment help) (cmd-parse-failure (error/whole-item-number :patch))
```
MeBrei: It's more and more difficult to keep an overview of this. Could the `patch-num-pat` or `_` commands somehow be extracted into one helper command?

jameshaydon: Let me try.

jameshaydon:
Closing because I realised `/cmd-n` throws and so making it higher order (take patters as arguments) is a bit of a rabbit hole. Will rethink cmd arg parsing later.

PR #565, opened 2019-03-14
Move to brew bundle

stephengroat:
`brew install` generates an error if the package is already installed and can create havoc for CI system when the base image changes
`brew bundle` is now included in homebrew and allows for textual configurations to be moved out of ci scripts and into the main repo for those developing on osx

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -53,10 +53,8 @@ jobs:
       - run:
           name: Homebrew Dependencies
           command: |
-            brew install coreutils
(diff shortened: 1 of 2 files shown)


xla:
@stephengroat Thanks for the contribution, it would be nice to get some context for why tihs is an improvement in the PR description so people get the context.

stephengroat:
@xla sorry about that, let me know if the edits work for you!

xla:
Great, thanks so much!

PR #572, opened 2019-03-15
List open/pending issues/patches by default.

jameshaydon:
- Listing issues and patches only shows open/pending items by default.
- Fix bug in options command parsing to allow choosing other than a superset of the default set.
- Summary of number of items by state is printed at the end of list.
- Fix bug in pattern matching where values would fail to match if they are the wrong type.

### DIFF
--- a/bin/rad-issue
+++ b/bin/rad-issue
@@ -6,7 +6,10 @@
 (load! (find-module-file! "prelude/io-utils.rad"))
 
 (file-module! "prelude/error-messages.rad")
+(file-module! "monadic/items.rad")
+
 (import prelude/error-messages :as 'error)
+(import monadic/items :as 'items)
 
 (import prelude/validation :as 'validation)
 (import prelude/machine :as 'machine)
@@ -193,6 +196,8 @@
           [(pretty-headline i) i]
           [(pretty-row i) i])))
 
+    (def issues (values (list-issues machine)))
+
     (def filt
       (match (map (fn [opt] (read (string-append ":" opt))) (lookup :state options))
         []      (fn [x] #t)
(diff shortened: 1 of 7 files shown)


bin/rad-issue
```diff
@@ -231,7 +236,8 @@
                _ (put-str! "No selection made."))
         (do
           (put-str! (pretty-table-header))
-          (map (fn [d] (put-str! (first d))) items))))))
+          (map (fn [d] (put-str! (first d))) items)
+          (put-str! (items/summary "issues" issues)))))))
```
MeBrei: Could the list function be completely moved to `items.rad` It would be nice in general to extract more code that is shared by both apps and possibly used by new ones too.

rad/prelude/dict.rad
```diff
@@ -76,3 +76,21 @@
 (:test "lookup-default"
   [ (lookup-default :one 2 {:one 1}) ==> 1 ]
   [ (lookup-default :two 2 {:one 1}) ==> 2 ])
+
+(def group-by
+  "Partitions the values in a sequence `xs`. The partitions are returns in a dict
+  keyed by the return value of `f`."
```
MeBrei: `are returned` ?

bin/rad-issue
jameshaydon: I will do it in a separate PR.

jameshaydon: #12 in Radicle issues

MeBrei:
The help text should reflect the default as well.
Documentation on `radicle.xyz` should also be changed once this is merged. Especially updating the help text in the cli interface and maybe also the sample output of the list commands in the tutorials.

PR #573, opened 2019-03-15
Prevent race conditions when loading machines

geigerzaehler:
This commit prevents race conditions when initiating two concurrent queries two an machine that is not yet loaded. Concurrent queries would lead to concurrent calls to `ensureMachineLoaded`. Before both calls would receive `Nothing` from `lookupMachine` and would start the loading process. We prevent this with the following changes

* `ensureMachineLoaded` uses `modifyMachine'` so that invocations are serialized.
* `loadMachine` does not insert the machine into the store
* `refreshAsReader` has been renamed `pullStoreInputs`. The function is extended to writer machines in which case it is a no-op

### DIFF
--- a/src/Radicle/Daemon/MachineStore.hs
+++ b/src/Radicle/Daemon/MachineStore.hs
@@ -12,6 +12,7 @@ module Radicle.Daemon.MachineStore
   , CachedMachines(..)
   , MonadMachineStore(..)
   , modifyMachine
+  , modifyMachine'
   , insertMachine
   , insertUninitializedReader
   , emptyMachine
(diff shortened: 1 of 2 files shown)


src/Radicle/Daemon/MachineStore.hs
```diff
@@ -99,6 +100,11 @@ modifyMachine id f = do
             (m', a) <- f m
             pure (Cached m', a)
 
+modifyMachine' :: forall a m. (MonadMachineStore m) => MachineId -> (Maybe CachedMachine -> m (Maybe CachedMachine, a)) -> m a
+modifyMachine' id f = do
+    machines <- askMachines
+    CMap.modifyValue id (getMachines machines) f
```
jameshaydon: Would it be possible to not use this function? I'm asking because @jkarni mentioned yesterday that we might want the daemon to stop following machines after a while. I *think* we can implement deleting items in `CMap` if we simplify the API to just `insert`, `delete` and `modifyExisting` which takes as argument a function `a -> m a`, that is, can't delete depending on the value it finds.

geigerzaehler: I think the generalized `modifyValue` is the only way we can prevent these race conditions. Using `insert` and potentially `delete` does not guarantee that request will be linearized.

PR #581, opened 2019-03-20
Add e2e patch tests for unauthorized commands

MeBrei:


test/e2e/PatchAppTest.hs
```diff
@@ -92,3 +93,21 @@ test_patch_accept = testCaseSteps "patch accept" $ \step -> do
...
+test_patch_unauthorized :: TestTree
+test_patch_unauthorized = testCaseSteps "patch unauthorized actions" $ \step -> do
+    step "init project with patch (maintainer)"
+    initProjectWithPatch
+    _ <- runTestCommand "git" ["checkout", "master"]
+
+    step "rad patch reject fails (contributer)"
+    _ <- runTestCommand "rad-key" ["create", "--force"]
+    rejectOutput <- runTestCommandForError "rad-patch" ["reject", "0"]
+    assertContains rejectOutput "Are you authorized to do so?"
```
jameshaydon: I think it would be less brittle to check for `"error"`. Unrelated to this PR, but I note that in `error-messages.rad`, some messages are prefixed with `error:`, some with `Error:` and some not prefixed at all. Maybe this should be made consistent.

MeBrei: :+1: I can create another issue for that

test/e2e/Test/E2ESupport.hs
```diff
@@ -154,5 +149,35 @@ runTestCommand' bin args inputLines = do
...
                 <> "-------------------\n"
 
+-- | Like 'runTestCommandForError'' but does not provide input to stdin.
+runTestCommandForError :: (HasCallStack) => FilePath -> [Text] -> TestM Text
+runTestCommandForError bin args = runTestCommandForError' bin args []
+
+-- | Like 'runTestCommand'' but expects the command to return an error.
+--
+-- If the command exits with a non-zero exit code, an exception is thrown.
+-- Otherwise stdout of the error is returned.
```
geigerzaehler:
```suggestion
-- Otherwise stdout of the command is returned.
```

test/e2e/Test/E2ESupport.hs
```diff
@@ -154,5 +149,35 @@ runTestCommand' bin args inputLines = do
...
+    (exitCode, out, _) <- executeCommand bin argsString inputLines
+    case exitCode of
+        ExitFailure _ -> pure $ toS $ trimNewline out
+        ExitSuccess ->
+            let commandLine = bin <> " " <> unwords argsString
+            in liftIO $ HUnit.assertFailure $
+                "Command " <> commandLine <> " succeeded,\n"
+                <> "but was expected to fail."
+
+executeCommand :: FilePath -> [String] -> [Text] -> TestM (ExitCode, String, String)
```
jameshaydon: Can you change the signature so that `argsString :: [Text]` to remove the duplicate `let argsString = map toS args` at the call sites.

test/e2e/Test/E2ESupport.hs
```diff
@@ -154,5 +149,35 @@ runTestCommand' bin args inputLines = do
...
+runTestCommandForError' bin args inputLines = do
+    let argsString = map toS args
+    (exitCode, out, _) <- executeCommand bin argsString inputLines
+    case exitCode of
+        ExitFailure _ -> pure $ toS $ trimNewline out
+        ExitSuccess ->
+            let commandLine = bin <> " " <> unwords argsString
+            in liftIO $ HUnit.assertFailure $
+                "Command " <> commandLine <> " succeeded,\n"
+                <> "but was expected to fail."
```
geigerzaehler: I think printing `stdout` and `stderr` would be helpful here, too.

PR #583, opened 2019-03-21
Change Generic FromToRad for single constructor datatypes

jameshaydon:
For single-constructor Haskell data types:
- `toRad` will skip the constructor,
- `fromRad` will optionally parse the constructor.

Note that while this change is backwards compatible, it is not forwards compatible: the public daemon will need to be updated before it can accept expressions created with this version.

### DIFF
--- a/src/Radicle/Internal/Core.hs
+++ b/src/Radicle/Internal/Core.hs
@@ -788,21 +788,32 @@ evenArgs name = \case
 -- * Generic encoding/decoding of Radicle values.
 
 toRadG :: forall a t. (HasEot a, ToRadG t (Eot a)) => a -> Annotated t ValueF
-toRadG x = toRadConss (constructors (datatype (Proxy :: Proxy a))) (toEot x)
+toRadG x = toRadConss (case conss of {[_] -> SingleCons; _ -> NonSingleCons})
+                      conss
+                      (toEot x)
+  where
+    conss = constructors (datatype (Proxy :: Proxy a))
+
+data ConstructorMultiplicity = SingleCons | NonSingleCons
 
 class ToRadG t a where
-  toRadConss :: [Constructor] -> a -> Annotated t ValueF
+  toRadConss :: ConstructorMultiplicity -> [Constructor] -> a -> Annotated t ValueF
+
+consArgs :: (CPA t , ToRadFields t a) => Fields -> a -> [Annotated t ValueF]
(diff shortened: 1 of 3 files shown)


src/Radicle/Internal/Core.hs
```diff
@@ -785,21 +785,28 @@ evenArgs name = \case
...
 
 toRadG :: forall a t. (HasEot a, ToRadG t (Eot a)) => a -> Annotated t ValueF
-toRadG x = toRadConss (constructors (datatype (Proxy :: Proxy a))) (toEot x)
+toRadG x = toRadConss (case conss of {[_] -> True; _ -> False}) conss (toEot x)
+  where
+    conss = constructors (datatype (Proxy :: Proxy a))
 
 class ToRadG t a where
-  toRadConss :: [Constructor] -> a -> Annotated t ValueF
+  toRadConss :: Bool -> [Constructor] -> a -> Annotated t ValueF
```
jkarni: Can you document what the `Bool` stands for (or use a new datatype)?

test/spec/Radicle/Tests.hs
```diff
@@ -785,6 +785,22 @@ test_from_to_radicle =
         [ testForType (Proxy :: Proxy [Text]) ]
     , testGroup "Generic a => a"
         [ testForType (Proxy :: Proxy Foo) ]
+    , testGroup "Single constructor with selectors"
+        [ testForType (Proxy :: Proxy Bar) ]
+    , testGroup "Single constructor with no selectors"
+        [ testForType (Proxy :: Proxy Baz) ]
+    , testGroup "Constructors optional for single constructor datatypes"
+        [ let v = Vec $ fromList [ kw "bar"
```
jkarni: Would it not make more sense to have:

```
Bar "hello" 42 == { :constructor :bar :value { :bar1 "hello" :bar2 42 }
```

jameshaydon: That's departing a lot from what it is currently, and so would make it very hard to be backwards compatible. This test is simply stating that what we _currently_ have still works.

jameshaydon: Also I think we discussed this in the PR which added FromToRad in the first place; that `[:just 1]` and `:nothing` was a lot nicer that `{:constructor :just :value 1}` and `{:constructor :nothing}`.

PR #589, opened 2019-03-26
Enable CORS for public frontends.

jameshaydon:
So that browser frontends can use the `query` endpoint of daemons.

### DIFF
--- a/exe/Daemon.hs
+++ b/exe/Daemon.hs
@@ -23,6 +23,7 @@ import qualified Data.Map.Strict as Map
 import qualified Data.Text as T
 import qualified Data.Time.Clock.System as Time
 import           Network.Wai.Handler.Warp (run)
+import           Network.Wai.Middleware.Cors (simpleCors)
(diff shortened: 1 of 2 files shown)


exe/Daemon.hs
```diff
@@ -110,6 +111,12 @@ main = do
           Right (Right ()) -> do
             logError' "Server stopped (this should not happen)" []
             exitFailure
+  where
+    -- | Allow Content-Type header with values other then allowed by simpleCors.
+    corsWithContentType = Cors.cors (const $ Just policy)
+      where
+        policy = Cors.simpleCorsResourcePolicy
```
jkarni: There might be potential problems down the line here, but they're not relevant while there is no authentication. Can you open an issue for authentication on the web ui, and mention taking a deeper look at CORS?

jameshaydon: https://github.com/radicle-dev/rad-issues-frontend/issues/1

PR #594, opened 2019-03-27
Extract help commands to cmd-parsing

MeBrei:


rad/prelude/cmd-parsing.rad
```diff
@@ -186,6 +186,22 @@
   [ (/cmd ["cmd-1"]) ==> :nothing]
   [ (catch-for-testing /cmd ["cmd-0" 1]) ==> "Too many arguments for command \"cmd-0\" help"])
 
+(def /cmd-help
+  "Matches a help command. Additional arguments are ignored."
+  (fn []
+    (fn [value]
+      (def help-cmds ["help" "-h" "--help"])
+      (if (elem? (first value) help-cmds)
```
jameshaydon: This throws when the args list is empty rather than returning `:nothing`. This would prevent the empty args pattern in `rad-replicate` from working.

PR #595, opened 2019-03-27
Extract verify item number from rad issue and patch

MeBrei:


rad/monadic/items.rad
```diff
@@ -218,3 +223,13 @@
           (put-str! (pretty-table-header))
           (map (fn [d] (put-str! (first d))) filtered-items)
           (put-str! (summary (items-name item-flag) items)))))))
+
+(def verify-item-number
+  "Verifies that the item corresponding to `item-number` is part of the given
+  list of `items`. Returns an error and exits if the item is not found."
+  (fn [item-number items item-flag]
+    (match (lookup-default item-number #f items)
```
jameshaydon: I think it would be better to use `lookup-maybe` here.

PR #596, opened 2019-03-27
Add missing line break in project init prompt

MeBrei:
Adds a line break after the prompt asking for the repo kind to set up a new project.

### DIFF
--- a/bin/rad-project
+++ b/bin/rad-project
@@ -122,7 +122,7 @@
                'url ["3. Remote origin of existing current git repository:"
                      (string-append "   " url)])
          ])))
-    (match (prompt! question)
+    (match (prompt! (string-append question "\n"))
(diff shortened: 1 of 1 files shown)


bin/rad-project
```diff
@@ -118,9 +118,9 @@
           "1. New peer-to-peer repository (git-ipfs **EXPERIMENTAL**)"
           "2. Add your own remote (e.g.: github / gitlab / ...)"]
          (match current-remote
-               "" []
+               "" [""]
                'url ["3. Remote origin of existing current git repository:"
-                     (string-append "   " url)])
+                     (string-append "   " url "\n")])
          ])))
     (match (prompt! question)
```
jameshaydon: Maybe it would be clearer to just replace `question` here with `(string-append question "\n")`.

Issue #29, opened 2018-07-02
List as function/lambda argument


jkarni:
Currently a lambda has to be explicit about the arguments:

```
(lambda (x y) (+ x y))
```

Unparenthesized lambda arguments should also be allowed, in which case the lambda receives a list as argument (where the list contains all arguments actually given). This allows varargs, and kwargs.

jameshaydon:
Done in #606

PR #606, opened 2019-04-01
Issue 29 - support lambda definition with an atom representing an arguments vector

maxhallinan:
Is there still interest in supporting a lambda arguments list as described on [issue 29](https://github.com/radicle-dev/radicle/issues/29)? If there is, what do you think of this approach? 

A few notes:

- This binds a vector to the atom instead of a list to maintain consistency with the current lambda syntax
- The generated call example is `(foo arg1 ...)`
- I haven't added tests yet

### DIFF
--- a/src/Radicle/Internal/Eval.hs
+++ b/src/Radicle/Internal/Eval.hs
@@ -57,17 +57,17 @@ specialForms :: forall m. (Monad m) => Map Ident ([Value] -> Lang m Value)
 specialForms = Map.fromList $ first Ident <$>
   [ ( "fn"
     , \case
-          args : b : bs ->
+          args : b : bs -> do
+            e <- gets bindingsEnv
+            let toLambda argNames = pure (Lambda argNames (b :| bs) e)
             case args of
               Vec atoms_ -> do
                 atoms <- traverse isAtom (toList atoms_) ?? toLangError (SpecialForm "fn" "One of the arguments was not an atom")
-                e <- gets bindingsEnv
-                pure (Lambda (PosArgs atoms) (b :| bs) e)
+                toLambda (PosArgs atoms)
               Atom name -> do
-                e <- gets bindingsEnv
-                pure (Lambda (VarArgs name) (b :| bs) e)
-              _ -> throwErrorHere $ SpecialForm "fn" "First argument must be a vector of argument atoms"
-          _ -> throwErrorHere $ SpecialForm "fn" "Need an argument vector and a body"
+                toLambda (VarArgs name)
+              _ -> throwErrorHere $ SpecialForm "fn" "Function parameters must be one atom or a vector of atoms"
+          _ -> throwErrorHere $ SpecialForm "fn" "Function needs parameters (one atom or a vector of atoms) and a body"
       )
   , ( "module", createModule )
   , ("quote", \case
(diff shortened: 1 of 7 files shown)


jameshaydon:
Hi @maxhallinan 

I think there is still interest for this but we had in mind maybe something a lot more general (the issue should have been updated):

The proposal is to make all binding locations, including in `fn` and `def`, pattern-matching locations. This would mean that the following would work:

```
(def '[x y] [1 2])

(def 'foo
  (fn '[x y] (+ x y))

(def 'bar
  (fn 'args (map f args)))
```

This is obviously a very breaking change since all variables in binding locations would now be quoted. This might be quite a lot of noise compared with how useful it is.

However the change in this PR seems like it would be fully backwards compatible so we can probably merge it quite quickly.

maxhallinan:
This sounds nice. So then every function is variadic? Or you require as many arguments as there are explicit bindings? 

No worries if you want to hold off entirely. I'm happy to close the PR. In terms of backwards compatibility, the atom in the arg list position is not quoted: `(fn foo args (first args))`. If this proposal requires that atom to be quoted, the syntax introduced here will be broken.

jameshaydon:
@maxhallinan 

Let's go ahead with this PR.

> This binds a vector to the atom instead of a list 
agreed.

> The generated call example is (foo arg1 ...)
I think this is fine.

Let me know when you want a review!

maxhallinan:
@jameshaydon I added a test. This is ready for review. Thanks in advance!

src/Radicle/Internal/Pretty.hs
```diff
@@ -74,13 +74,21 @@ instance forall t. (Copointed t, Ann.Annotation t) => PrettyV (Ann.Annotated t V
...
         escapeStr = T.init . T.tail . show
-        prettyLambda ids vals =
+        prettyLambda (PosArgs ids) vals =
           let args :: Ann.Annotated t ValueF = Vec $ Seq.fromList (Atom <$> ids)
           in parens $
                annotate TAtom "fn" <+>
                align (sep [ prettyV args
                           , sep $ prettyV <$> toList vals
                           ])
+        prettyLambda (VarArgs id) vals =
```
jameshaydon: Remove duplication with the case above, move the case analysis into the `let args = ...`.

src/Radicle/Internal/Eval.hs
```diff
@@ -222,11 +232,14 @@ callFn f arguments = case f of
     evalManyWithEnv :: Env Value -> NonEmpty Value -> Lang m Value
     evalManyWithEnv env exprs =
         NonEmpty.last <$> withEnv (const env) (traverse baseEval exprs)
-    argumentBindings :: [Ident] -> Lang m (Env Value)
-    argumentBindings names =
+    posArgBindings :: [Ident] -> Lang m (Env Value)
```
jameshaydon: To reduce the number of cases above, I would keep this as one function `argumentBindings :: LambdaArgs -> Lang m (Env Value)`.

src/Radicle/Internal/Eval.hs
```diff
@@ -62,7 +62,10 @@ specialForms = Map.fromList $ first Ident <$>
               Vec atoms_ -> do
                 atoms <- traverse isAtom (toList atoms_) ?? toLangError (SpecialForm "fn" "One of the arguments was not an atom")
                 e <- gets bindingsEnv
-                pure (Lambda atoms (b :| bs) e)
+                pure (Lambda (PosArgs atoms) (b :| bs) e)
+              Atom name -> do
+                e <- gets bindingsEnv
+                pure (Lambda (VarArgs name) (b :| bs) e)
               _ -> throwErrorHere $ SpecialForm "fn" "First argument must be a vector of argument atoms"
```
jameshaydon: These two error messages need to be updated.

src/Radicle/Internal/Eval.hs
```diff
@@ -62,7 +62,10 @@ specialForms = Map.fromList $ first Ident <$>
               Vec atoms_ -> do
                 atoms <- traverse isAtom (toList atoms_) ?? toLangError (SpecialForm "fn" "One of the arguments was not an atom")
                 e <- gets bindingsEnv
-                pure (Lambda atoms (b :| bs) e)
+                pure (Lambda (PosArgs atoms) (b :| bs) e)
+              Atom name -> do
+                e <- gets bindingsEnv
```
jameshaydon: The `e <- gets bindingsEnv` can be moved above the case analysis on `args`.

src/Radicle/Internal/Pretty.hs
maxhallinan: 👍🏻updated

src/Radicle/Internal/Eval.hs
maxhallinan: Great suggestion. Updated.

src/Radicle/Internal/Eval.hs
maxhallinan: Let me know if the new wording here is confusing.

src/Radicle/Internal/Eval.hs
maxhallinan: 👍🏻updated

maxhallinan:
@jameshaydon thanks for the review! I updated with your suggestions.

PR #610, opened 2019-04-02
Show call example in doc function

MeBrei:
For now this will only show the call example for non primfns

### DIFF
--- a/src/Radicle/Internal/Core.hs
+++ b/src/Radicle/Internal/Core.hs
@@ -18,6 +18,7 @@ import           Data.Scientific (Scientific)
 import           Data.Semigroup ((<>))
 import           Data.Sequence (Seq(..))
 import qualified Data.Sequence as Seq
+import qualified Data.Text as T
 import           Generics.Eot
 import qualified GHC.Exts as GhcExts
 import qualified GHC.IO.Handle as Handle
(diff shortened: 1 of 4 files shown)


src/Radicle/Internal/Core.hs
```diff
@@ -535,10 +536,22 @@ addBinding :: Ident -> Maybe Text -> Value -> Bindings m -> Bindings m
...
 lookupAtomWithDoc :: Monad m => Ident -> Lang m (Doc.Docd Value)
-lookupAtomWithDoc i = get >>= \e -> case Map.lookup i . fromEnv $ bindingsEnv e of
+lookupAtomWithDoc i@(Ident name) = get >>= \e -> case Map.lookup i . fromEnv $ bindingsEnv e of
     Nothing -> throwErrorHere $ UnknownIdentifier i
-    Just x  -> pure x
+    Just x@(Doc.Docd d_ v)  -> case callExample name v of
+      Nothing -> pure x
+      Just tt -> maybe (pure x) (doc tt v) d_
+  where
+    doc t v d = pure (Doc.Docd (Just (t <> "\n\n" <> d)) v)
```
MeBrei: I could add another layer of `maybe` here, but it might make it less clear to read.

jameshaydon: If there is no documentation-string do we still want to return the call-example?

If so `callExample name v` and `d_` are both `Maybe Text`, and what you want to do is `<>` them with `"\n\n"` in-between, so you can use `<>` directly because `Maybe a` is a semigroup when `a` is.

If not, then you want to preserve the `Maybe`-ness of `d_`, so you `fmap` over it with `(maybe "" (<> "\n\n") (callExample name v) <>)`.

Personally, I think it probably makes sense to return the call example.

MeBrei: Currently `"No documentation found for and."` is shown when the doc is missing. Should it then show the call-example plus this note? 

On the other hand updating the reference document fails if the doc string is missing, so maybe we don't need to plan for the case of a missing doc string in the repl

jameshaydon: The reason a missing docstring causes a failure when generating the reference doc is for a different reason: that's for the CI to force us to document all functions of the prelude.

I think it would be useful to always get the call-example at the REPL, but your call.

PR #612, opened 2019-04-02
Update docs

### LINKED ISSUE #608: Language docs improvements
- [x] Link to `docs.radicle.xyz` from `radicle.xyz`
- [x] Finish or remove chain example from docs
- [x] Fix indexing in docs
- [x] Mention that `;;` is for comments early on, and what `expr ;; ==> val` means
- [x] Note in docs.radicle.xyz getting started that after `stack install` you might need to change your path

jameshaydon:
Fixes #608, also adds documentation for keywords and atoms to the guide.

The radicle docs are linked from radicle.xyz, but it's not very visible (in http://radicle.xyz/docs/index.html#radicle-p2p-architecture).


docs/source/guide/Basics.lrad
```diff
@@ -1,16 +1,28 @@
...
+[Clojure](clojure.org/). To effectively use it, it is helpful to learn the
+basics of it independent of its application.
 
 This chapter introduces `radicle` as a language. In each section we present
 a core datatype of the language, and show some basic operations that can be
 performed on them.
 
+## Comments
+
+The character `;` starts a comment. This comments all the text until the next
```
jkarni: It's the string `;;` that srtas a comment.

jkarni:
```suggestion
The string `;;` starts a comment. This comments all the text until the next
```

docs/source/guide/Basics.lrad
```diff
@@ -32,6 +44,30 @@ Strings are enclosed in double quotes:
 "this is a string"
 ```
 
+### Keywords
+
+Keywords are identifiers that evaluate to themselves. To create a keyword you
+prefix `:` onto an identifier, for example `:foo` is a keyword. Keywords are
```
jkarni:
```suggestion
prefix `:` onto an identifier (`:foo` is a keyword, for example). Keywords are
```

docs/source/guide/Basics.lrad
jameshaydon: Oh... but for every other lisp it's `;` (Scheme, Clojure, etc.). Can I fix the parser rather than the docs?

jkarni: Up to you. On the one hand consistency with other lisps would be nice. On the other, we can't then use `;` in code...

docs/source/guide/Issues.lrad
```diff
@@ -1,4 +1,4 @@
-# Issues machine
+# Issues machine tutorial
```
MeBrei: Maybe `Issue machine` as it is also singular in `rad issue` and `issue-remote.rad`

PR #614, opened 2019-04-04
Make 'repl' a rad command.

jkarni:
Symlinks the binary into the old location so find-module-file still
finds it.


bin/rad-repl
```diff
@@ -0,0 +1,23 @@
+#!/usr/bin/env radicle
+(load! (find-module-file! "prelude.rad"))
+(import prelude/machine '[eval] :unqualified)
+
+(def help
+  "Default help text."
+  (fn []
+    "radicle is a LISP-like language intended for programming chains, and interacting with those chains.
+
+    Type (doc '<name>) for further documentation of <name>."))
```
MeBrei: `doc!` is nice to read especially if #610 is merged.

bin/rad-repl
```diff
@@ -0,0 +1,23 @@
...
+
+(def read-line-or-exit! (fn []
```
MeBrei: line break missing

bin/rad-repl
```diff
@@ -0,0 +1,23 @@
+#!/usr/bin/env radicle
+(load! (find-module-file! "prelude.rad"))
+(import prelude/machine '[eval] :unqualified)
+
+(def help
+  "Default help text."
+  (fn []
+    "radicle is a LISP-like language intended for programming chains, and interacting with those chains.
```
MeBrei: Wouldn't it be nicer if help would directly print the help text ? easier to read then.

jkarni:
> Is there not some way to reduce the code duplication with rad/repl.rad?

Hm, I thought I'd symlinked them. Might have been a hardlink that git dereferenced. Let me try again.

jkarni:
@jameshaydon suggested having rad/repl.rad be a module, and bin/rad-repl import that and run the repl function, so that we can have easy REPLs for specific machines (e.g. `rad issue repl`). Going to do that.

jkarni:
@jameshaydon @MeBrei wanna review again?

Issue #587, opened 2019-03-22
`Could not connect to the Radicle daemon` on `rad project checkout ...` with both daemons running


hxrts:
reported by [@Bag_End_](https://twitter.com/Bag_End_) on twitter

## system details

- Ubuntu installed via apt
- not behind firewall
- IPFS connects when run for non-radicle purposes

## terminal output

`rad daemon-ipfs` output:

```
Initializing daemon... go-ipfs version: 0.4.18- Repo version: 7 System version: amd64/linux Golang version: go1.11.1 Successfully raised file descriptor limit to 2048. Swarm is limited to private network of peers with the swarm key Swarm key fingerprint: 9bbc6ea04403a4595876eedfadb1aa93 Swarm listening on /ip4/127.0.0.1/tcp/9303 Swarm listening on /ip4/172.17.0.1/tcp/9303 Swarm listening on /ip4/172.18.0.1/tcp/9303 Swarm listening on /ip4/172.19.0.1/tcp/9303 Swarm listening on /ip4/172.20.0.1/tcp/9303 Swarm listening on /ip4/192.168.1.18/tcp/9303 Swarm listening on /ip6/::1/tcp/9303 Swarm listening on /p2p-circuit Swarm announcing /ip4/127.0.0.1/tcp/9303 Swarm announcing /ip4/172.17.0.1/tcp/9303 Swarm announcing /ip4/172.18.0.1/tcp/9303 Swarm announcing /ip4/172.19.0.1/tcp/9303 Swarm announcing /ip4/172.20.0.1/tcp/9303 Swarm announcing /ip4/192.168.1.18/tcp/9303 Swarm announcing /ip6/::1/tcp/9303 API server listening on /ip4/127.0.0.1/tcp/9301 Gateway (readonly) server listening on /ip4/127.0.0.1/tcp/9302 Daemon is ready
```

`rad daemon-radicle` output:

```
INFO    Start listening port=8909
INFO    Following as reader machine-id=12D3KooWRsrUcqCFU3ompktgTo197gguibN3cjmgaoQy8AS54iHn current-input-index=MachineEntryIndex zdpuB1dS7MzsbQnZ7hM1CqoQKG6eXcjmjHUh3VqFKnP3CRf4c
```

`rad project checkout 12D3KooWRsrUcqCFU3ompktgTo197gguibN3cjmgaoQy8AS54iHn` output:

```
Could not connect to the Radicle daemon, most likely it is not running. Please see the documentation for how to start it.
[...]

PR #620, opened 2019-04-11
Fix broken pubsub subscribing.

jkarni:
Fixes the other half of the problems people were seeing in #587

### DIFF
--- a/src/Radicle/Ipfs.hs
+++ b/src/Radicle/Ipfs.hs
@@ -214,11 +214,13 @@ subscribe :: (MonadIpfs m) => Text -> (PubsubMessage -> IO ()) -> m ()
 subscribe topic messageHandler = do
     Client{clientHttpManager, clientBaseRequest} <- askClient
     liftIO $ runResourceT $ do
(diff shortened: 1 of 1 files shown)


src/Radicle/Ipfs.hs
```diff
@@ -214,11 +214,13 @@ subscribe :: (MonadIpfs m) => Text -> (PubsubMessage -> IO ()) -> m ()
...
     liftIO $ runResourceT $ do
-        let req = clientBaseRequest
-                & HttpClient.setQueryString
-                    [ ("arg", Just $ T.encodeUtf8 topic)
-                    , ("encoding", Just "json")
-                    , ("stream-channels", Just "true")
+        let req = makeIpfsRequest
+                    clientBaseRequest
+                    Http.methodGet
+                    "pubsub/sub"
```
geigerzaehler: Am I reading this right that before we just made a request to the root URL and this never actually worked?

jkarni: Not in the past 20 days...

PR #621, opened 2019-04-11
Some follow-up on #617.

jameshaydon:


src/Radicle/Internal/ConcurrentMap.hs
```diff
@@ -26,28 +26,26 @@ empty = CMap <$> newMVar Map.empty
 
 -- | Atomically lookup a value.
 lookup :: Ord k => k -> CMap k v -> IO (Maybe v)
-lookup k (CMap m_) = withMVar m_ $ \m -> do
-   case Map.lookup k m of
-     Nothing -> pure Nothing
-     Just v_ -> readMVar v_
+lookup k m = modifyValue k m (\x -> pure (x,x))
```
MeBrei: If `Map.lookup k m` returns `Nothing` wont this now lead to it being inserted in `modifyValue`?

jameshaydon: I'm not sure what you mean. Are you commenting on the old code or the new?

MeBrei: On the new... In the old code, `Map.lookup k m` in `lookup` returned `Nothing` when the key didn't exists, now, as it uses `modifyValue`, `Map.lookup k m` will return `pure (Map.insert k valueVar m, valueVar)`

jkarni: Yeah, it does seem to me like @MeBrei is right and this leads to `Nothing`s being inserted.

jameshaydon: Ah I see it adds useless `Nothing`s, agreed.

jameshaydon:
I'm going to wait for the improved e2e before merging.

PR #629, opened 2019-04-26
Allow unsigned request

MeBrei:
This is a quick way of allowing unsigned requests. 
Setting up a new machine with the updated validation, I was able to send an unsigned request via jsradicle. The missing `:author` is just shown as `???`. If the request is limited to admins or a specific author, the validations fail as expected.

The current implementation has no "nice" validation failure message if there is no signature and the author is not `:browser`. The change automatically applies to all machines. It could be separated out to allow unsigned requests only in specific places.

Unsigned commands can be tested with `SendUnsignedCommand` from https://gist.github.com/MeBrei/1b998ac354b19ed7465903cb3df21e13

### DIFF
--- a/rad/monadic/issue-remote.rad
+++ b/rad/monadic/issue-remote.rad
@@ -183,10 +183,8 @@
       x
       (insert :modified-at (lookup :created-at x) x))))
 
-(def create-issue
-  "Create an issue from a dict, checking that it is valid and storing it in `issues`."
+(def add-issue
   (fn [i]
-    (validator/input i)
     (def fixed
       (insert :comments
               (map fix-modified-at (lookup :comments i))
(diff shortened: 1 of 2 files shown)


MeBrei:
To test replace `run` of `jsradicle` with:
```
const run = async () => {
  var machine = <issue-machine-id>
  var machineId = "monadic/radicle/issue"
  var payload = {":labels": [], ":comments": [], ":state": ":open", ":body": "testing", ":title": "title testing", ":created-at": "2019-04-26T08:18:48Z", ":modified-at": "2019-04-26T08:18:48Z"};
  
  var res = sendUnsignedCommand(machine, machineId, "anonymous-create-issue", bla);
  res.then(function (response) {
    console.log(response);
  })
  .catch(function (error) {
    console.log(error);
  });
}
```

PR #630, opened 2019-04-26
feat: Machines define their frontend

jkarni:
Creates a new endpoint for the daemon, at the top-level for the machine, which responds with HTML.

In order to make it easy to deploy a frontend that may contain multiple files, without needing to inline resources, the idea here is that we use an entire dir instead of a single file. So when you visit `http://localhost:8909/v0/machines/<machineid>/frontend/index.html`, the machine is queried for `(get-html)`, which results in a `CID1`, and then IPFS is queried for `CID1/index.html`. Similarly, if you query `http://localhost:8909/v0/machines/<machineid>/frontend/somepath`, `CID1/somepath` is queried.

The upside, as mentioned, is really easy deploys. Just do `rad ipfs add -w -r *` in your build dir, and then add the resulting CID as the value returned by `get-html` (as long as your relative links use `.` as a root). The downside is that assets are not served by (ipfs) name, so we can't do loads of caching.


exe/Daemon.hs
```diff
@@ -238,6 +239,28 @@ send id (Api.SendRequest expressions) = do
         Nothing -> throw $ MachineError id AckTimeout
 
 
+-- | Give the HTML specified by a machine (via defining `(get-html)`).
+frontend :: MachineId -> [Text] -> Daemon Api.HtmlText
+frontend id path = do
+    m <- ensureMachineLoaded id
+    bumpPolling id
+    case runIdentity $ interpret "[frontend]" "(get-html)" (machineState m) of
+        Left err -> throw $ MachineError id (InvalidInput err)
```
jameshaydon: Maybe the error can be more informative, for example to include a message that the machine might not have defined a front-end at all, or not according to this convention.

exe/Daemon.hs
```diff
@@ -238,6 +239,28 @@ send id (Api.SendRequest expressions) = do
...
+                    (InvalidInput $ toLangError $ OtherError "get-html returned invalid CID")
+                Right v' -> do
+                    let ipfsPath = show v' <> "/" <> T.intercalate "/" path
+                    logInfo "Query HTML" [ ("CID", show v')
+                                         , ("machine-id", getMachineId id)
+                                         , ("ipfs-path", ipfsPath)
+                                         ]
+                    val <- Ipfs.ipfsHttpGet' "cat" [("arg", ipfsPath)]
+                    pure $ Api.HtmlText $ toS val
+        Right _ -> throw $ MachineError id (DaemonError "get-html returned non-string")
```
jameshaydon: I don't think the error message should assume `get-html` is a function.

jkarni: Fixed

exe/Daemon.hs
jkarni: Fixed

PR #638, opened 2019-05-09
New show command

MeBrei:
! This breaks existing machines ! 

This creates a new show command `show-unbound` which does not add tabs/line breaks.

This simplifies signing requests in other languages, fe jsradicle.

### DIFF
--- a/src/Radicle/Internal/Pretty.hs
+++ b/src/Radicle/Internal/Pretty.hs
@@ -164,6 +164,15 @@ renderCompactPretty = renderStrict . layoutCompact . prettyV
 renderPretty :: PrettyV v => PageWidth -> v -> Text
 renderPretty pg = renderStrict . layoutSmart (LayoutOptions pg) . prettyV
 
+-- | 'renderPretty', but with unbounded width.
+--
+-- Example:
+--
+-- >>> renderPrettyUnbounded (asValue (List [String "hi", String "there"]))
+-- "(\"hi\" \"there\")"
+renderPrettyUnbounded :: PrettyV v => v -> Text
+renderPrettyUnbounded = renderPretty Unbounded
+
 -- | 'renderPretty', but with default layout options (80 chars, 1.0 ribbon)
 --
 -- Examples:
--- a/src/Radicle/Internal/Pretty.hs
+++ b/src/Radicle/Internal/Pretty.hs
@@ -66,7 +66,7 @@ instance forall t. (Copointed t, Ann.Annotation t) => PrettyV (Ann.Annotated t V
         PrimFn i -> pretty i
         Dict mp -> braces . align $
             sep [ prettyV k <+> prettyV val
-                | (k, val) <- Map.toList mp ]
+                | (k, val) <- Map.toAscList mp ]
         Lambda ids vals _ -> prettyLambda ids vals
         LambdaRec _ ids vals _ -> prettyLambda ids vals
         VEnv _ -> angles "env"
--- a/src/Radicle/Internal/PrimFns.hs
+++ b/src/Radicle/Internal/PrimFns.hs
@@ -553,9 +553,9 @@ purePrimFns = fromList $ allDocs $
     , ( "show"
       , "Returns a string representing the argument value."
       , oneArg "show" (pure . String . renderPrettyDef))
-    , ( "show-unbound"
+    , ( "show-unformatted"
       , "Returns a string representing the argument value. (No extra formatting)"
-      , oneArg "show-unbound" (pure . String . renderPrettyUnbounded))
+      , oneArg "show-unformatted" (pure . String . renderPrettyUnbounded))
     , ( "seq"
       , "Given a structure `s`, returns a sequence. Lists and vectors are returned\
         \ without modification while for dicts a vector of key-value-pairs is returned:\
--- a/src/Radicle/Internal/PrimFns.hs
+++ b/src/Radicle/Internal/PrimFns.hs
@@ -555,7 +555,7 @@ purePrimFns = fromList $ allDocs $
       , oneArg "show" (pure . String . renderPrettyDef))
     , ( "show-unbound"
       , "Returns a string representing the argument value. (No extra formatting)"
-      , oneArg "show-com" (pure . String . renderPrettyUnbounded))
+      , oneArg "show-unbound" (pure . String . renderPrettyUnbounded))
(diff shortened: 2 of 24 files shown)


src/Radicle/Internal/PrimFns.hs
```diff
@@ -553,6 +553,9 @@ purePrimFns = fromList $ allDocs $
     , ( "show"
       , "Returns a string representing the argument value."
       , oneArg "show" (pure . String . renderPrettyDef))
+    , ( "show-unbound"
+      , "Returns a string representing the argument value. (No extra formatting)"
+      , oneArg "show-com" (pure . String . renderPrettyUnbounded))
```
```suggestion
      , oneArg "show-unbound" (pure . String . renderPrettyUnbounded))
```

src/Radicle/Internal/Pretty.hs
```diff
@@ -164,6 +164,15 @@ renderCompactPretty = renderStrict . layoutCompact . prettyV
...
 renderPretty pg = renderStrict . layoutSmart (LayoutOptions pg) . prettyV
 
+-- | 'renderPretty', but with unbounded width.
+--
+-- Example:
+--
+-- >>> renderPrettyUnbounded (asValue (List [String "hi", String "there"]))
+-- "(\"hi\" \"there\")"
+renderPrettyUnbounded :: PrettyV v => v -> Text
+renderPrettyUnbounded = renderPretty Unbounded
```
jkarni: Can you add a test somewhere that generates arbitrary values that don't include strings with whitespace, and then check that `rendePrettyUnbound` never results in multiple consecutive spaces, tabs, or newlines?

MeBrei: I added tests for `show-unbound` which inderectly tests `renderPrettyUnbounded`

test/spec/Radicle/Internal/Arbitrary.hs
```diff
@@ -70,3 +53,33 @@ instance Arbitrary a => Arbitrary (Bindings a) where
 
 instance Arbitrary a => Arbitrary (Doc.Docd a) where
     arbitrary = Doc.Docd Nothing <$> arbitrary
+
+gen :: Gen Text -> Int -> Gen Value
+gen textGen n | n == 0 = frequency $ first pred <$> freqs
```
jkarni: Can you give a more descriptive name to this, and maybe a docstring?

test/spec/Radicle/Internal/Arbitrary.hs
```diff
@@ -19,34 +22,14 @@ instance Arbitrary r => Arbitrary (Env r) where
...
-        prims = purePrimFns
-        isPrimop x = x `elem` Map.keys (getPrimFns prims)
-        isNum x = isJust (readMaybe (toS $ fromIdent x) :: Maybe Scientific)
-        lambdaArgs = oneof [ PosArgs <$> sizedList, VarArgs <$> arbitrary ]
+    arbitrary = sized (gen arbitrary)
+
+newtype NoSpaceValue = NoSpaceValue Value
+  deriving Show
+
+instance Arbitrary NoSpaceValue where
```
jkarni: Maybe rename this to `NoDoubleSpacesValue`?

test/spec/Radicle/Tests.hs
```diff
@@ -561,6 +571,10 @@ test_eval =
   where
     failsWith src err    = noStack (runPureCode src) @?= Left err
     succeedsWith src val = runPureCode src @?= Right val
+    no_tabs t = not $ T.any (== '\t') t
+    no_line_breaks t = not $ T.any (== '\n') t
+    no_double_spaces t = not $ T.isInfixOf "  " t
+    no_unallowed_whitespace t = no_double_spaces t && no_line_breaks t && no_tabs t
```
jkarni: Camel case.

test/spec/Radicle/Tests.hs
```diff
@@ -475,6 +475,16 @@ test_eval =
         runPureCode "(show (dict 'a 1))" @?= Right (String "{a 1}")
         runPureCode "(show (fn [x] x))" @?= Right (String "(fn [x] x)")
 
+    , testCase "'show-unbounds' works" $ do
+        runPureCode "(show-unbound {:b 2 :a 1})" @?= Right (String "{:a 1 :b 2}")
+        runPureCode "(show-unbound {:bar \"bar bar bar bar bar bar bar bar bar bar bar bar bar bar bar\" :foo \"foo foo foo foo foo\"})"
+          @?= Right (String "{:bar \"bar bar bar bar bar bar bar bar bar bar bar bar bar bar bar\" :foo \"foo foo foo foo foo\"}")
+
+    , testProperty "'show-unbounds does not add any whitespace characters'" $ \(NoSpaceValue val :: NoSpaceValue) -> do
+        let res = runPureCode $ toS [i|(show-unbound (quote #{renderPrettyUnbounded val}))|]
```
jkarni: Is the quote needed?

src/Radicle/Internal/PrimFns.hs
```diff
@@ -553,6 +553,9 @@ purePrimFns = fromList $ allDocs $
     , ( "show"
       , "Returns a string representing the argument value."
       , oneArg "show" (pure . String . renderPrettyDef))
+    , ( "show-unbound"
+      , "Returns a string representing the argument value. (No extra formatting)"
```
jkarni: Maybe `show-unformatted`?

test/spec/Radicle/Tests.hs
MeBrei: Should only be around the function name

jkarni: Not sure I understand.

Right now it seems like some pretty-printing is not being tested because of quote. E.g., lambdas.

MeBrei: Sorry, misread your message (which pointed me to a different typo)

The quote is needed as otherwise on show/show-unformatted unknown identifiers (that are arbitrarily created) lead to errors. Quoting prevents that. I don't see what the problem with lambdas is, at least from testing in repl.

jkarni:
```
rad> (type (quote (fn [] 3)))
:list
rad> (type (fn [] 3))
:function
```
So you're never giving `show-unbound` an argument that's a function, as far as I can tell.

You can just make the test assert that *if* the result is not an error, it satisfies your property (or else discard errors).

PR #641, opened 2019-05-14
Macros

jameshaydon:
Adds macros.

- Macros are *not* hygienic. Hygienic macros are:

  - complex to implement

  - hard to specify

  - have lots of weird corner cases

  (see e.g. http://okmij.org/ftp/Scheme/macros.html#syntax-rule-dark-corner)
  For these reasons I don't think they are a good idea for Radicle.

- Macros not being hygienic is okay because:
  - Macros can insert values directly into the returned expressions (that is, macro-expansion code may contain closures, etc.), and therefore can insert values from the definition site.
  - Macro-expanded code is evaluated in its own scope.
  - A `gensym` function is included.

- `if` is now a macro.

- `(:test ...)` is replaced by a `test` macro.

- `tx` functions no longer need the state argument, because tests aren't made using a a tx-function anymore.


rad/prelude/basic.rad
```diff
@@ -92,17 +98,31 @@
...
   [ (read-many "([:one] {:two 3}) (four \"five\" 6)") ==> [([:one] {:two 3}) (four "five" 6)]])
 
 (def <=
   "Test if `x` is less than or equal to `y`."
   (fn [x y] (or (eq? x y) (< x y))))
+
+(def gensym-counter (ref 0))
+
+(def gensym
+  "Returns a new symbol with a pseud-unique name. The returned symbols are `G__#`,
```
jkarni:
```suggestion
  "Returns a new symbol with a pseudo-unique name. The returned symbols are `G__#`,
```
Could you explain what `pseudo-unique` means?

rad/prelude/basic.rad
```diff
@@ -1,8 +1,14 @@
 {:module  'prelude/basic
  :doc     "Basic function used for checking equality, determining the type of a value, etc."
- :exports '[or some empty-seq? length maybe->>= maybe-foldlM elem? head tail read read-many <=]}
+ :exports '[if or some empty-seq? length maybe->>= maybe-foldlM elem? head tail read read-many <=]}
```
jkarni: Export `gensym`?

rad/prelude/machine-remote-test.rad
```diff
@@ -0,0 +1,53 @@
+{:module 'prelude/machine-remote-test
+ :doc "Tests for machine-remote."
+ :exports []}
+
+(test
+ "lock-machine"
+ [:setup
+  (do
```
jkarni: We no longer need a `do` block in `:setup`

rad/prelude/test-macro.rad
```diff
@@ -0,0 +1,46 @@
+{:module 'prelude/test-macro
+ :doc
+ "Provides eval that adds a `test` macro. If an the eval encounters an
+   expression of the form `(test \"desc\" tests...)` we add it to the vector
+   stored in the `tests` ref. The tests are then run by `rad/prelude/test.rad`."
+ :exports '[test tests]}
+
+(def tests
+  "Reference that collects tests defined with the `test` macro. This
+  value is updated by a custom `eval`."
```
jkarni: Do you mean `by the test macro`?

rad/prelude/test-macro.rad
```diff
@@ -0,0 +1,46 @@
...
+          (def y (ref 0))
+          (write-ref y 1))
+    ]
+    [(+ 3 2) ==> 5]
+    [x       ==> 3]
+)
+
+(test "'test' handles exceptions properly"
+    [:setup
+      (do
```
jkarni: No `do` block

rad/prelude/test-macro.rad
```diff
@@ -0,0 +1,46 @@
...
+                 (list add-right
+                       {:doc-test (list 'quote (drop 1 es))
+                        :name     (nth 0 es)
+                        :env      (list get-current-state)}
+                       (list read-ref tests)))
+           :test-added))))
+
+(test "'test' works (including :setup)"
+    [:setup
+       (do
```
jkarni: No `do` block

src/Radicle/Internal/Pretty.hs
```diff
@@ -64,6 +64,7 @@ instance forall t. (Copointed t, Ann.Annotation t) => PrettyV (Ann.Annotated t V
           v' :<| Empty -> brackets $ prettyV v'
           _            -> brackets . sep $ prettyV <$> toList vs
         PrimFn i -> pretty i
+        Macro _ -> angles "macro"
```
jkarni: We can't just show it?

src/Radicle/Internal/Core.hs
```diff
@@ -292,6 +293,7 @@ valType = \case
   ProcHandle _ -> TProcHandle
   Lambda{} -> TFunction
   LambdaRec{} -> TFunction
+  Macro{} -> TFunction -- debatable lol
```
jkarni: `TMacro`?

src/Radicle/Internal/Pretty.hs
jameshaydon: Oh I have some comments about this, see IRC.

rad/prelude/basic.rad
```diff
@@ -92,17 +99,40 @@
...
+
+(def gensym
+  "Returns a new symbol with a pseudo-unique name. The returned symbols are `G__#`,
+  where # is a number.
+
+  This is used for generating names for \"fresh\" variables in macro-expanded
+  code. Technically it is possible to still get a clash, but this won't happen
+  as long as:
+  
+  - This function is the only one that updates `gensym-counter`, which isn't
```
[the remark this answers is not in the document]
jameshaydon: I have hidden `gensym-counter` inside a closure instead.

PR #642, opened 2019-05-17
Drive-by: Attempt to fix #636

### LINKED ISSUE #636: Pubsub reconnect backoff
We use full-jitter (~ exponential) backoff for retries if a pubsub connection drops. That's nice, but if a connection then *succeeds* for a while (works for a few days), and only then fails again, the backoff is not reset. So long-running daemons will eventually just be at the max delay (1 min).

kim:


src/Radicle/Ipfs.hs
```diff
@@ -248,11 +251,14 @@ subscribe topic messageHandler = do
...
-            policy
+    withBackoff action = do
+        lastSuccess <- liftIO $ Clock.getTime Clock.Monotonic >>= newIORef
+        recovering
+            (resetAfter lastSuccess (60 * minutes) policy)
             (skipAsyncExceptions ++ [h]) -- `h` should be *after* skipAsync!
-            (const action)
+            . const $
+                action <* liftIO
+                    (Clock.getTime Clock.Monotonic >>= writeIORef lastSuccess)
```
jkarni: `action` never returns in this case (only throws), so it seems like the rhs of `<*` never gets executed? Maybe a catch that rethrows after setting the IORef?

(I'd also maybe rename `lastSuccess` to `lastFailure`)

kim: Ah true, brainfart.

src/Radicle/Ipfs.hs
```diff
@@ -42,11 +42,13 @@ module Radicle.Ipfs
 import           Protolude hiding (TypeError, catch, catches, try)
 
 import           Control.Exception.Safe
```
jkarni: Maybe make the unqualified exports explicit then?

kim: Like this ea987dc0bc21a296369e86fdba351501ca6ca90f?

I used the qualified `Safe.onException` to make the semantics wrt `skipAsyncExceptions` more explicit. If you're using `safe-exceptions` everywhere anyways, that's maybe not necessary.

PR #644, opened 2019-05-21
Make import fail for undefined symbols (fixes #631)

### LINKED ISSUE #631: Import should fail if a symbol isn't defined
At the moment `(import foo '[bar] :unqualified)` will succeed even if `bar` is not defined/exported by `foo`.

curiousleo:
Fixes #631.

### DIFF
--- a/bin/rad-project
+++ b/bin/rad-project
@@ -13,7 +13,7 @@
 (import prelude/cmd-parsing :unqualified)
 (import prelude/io :as 'io)
 (import prelude/io-utils
-  '[set-git-config get-git-config process-git-with-exit!]
+  '[set-git-config! get-git-config! process-git-with-exit!]
   :unqualified)
 
 (import monadic/issue '[create-issue-machine!] :unqualified)
--- a/src/Radicle/Internal/PrimFns.hs
+++ b/src/Radicle/Internal/PrimFns.hs
@@ -709,8 +709,10 @@ purePrimFns = fromList $ allDocs $
           let allMod = fromEnv e
           toImport <- case v_ of
                 Just vs -> do
-                  is <- Protolude.toList <$> traverse isAtom vs ?? toLangError (OtherError "import: must be given a vector of symbols to import")
-                  pure $ Map.restrictKeys allMod (Set.fromList is)
+                  is <- Set.fromList <$> Protolude.toList <$> traverse isAtom vs ?? toLangError (OtherError "import: must be given a vector of symbols to import")
+                  if Set.null (is Set.\\ Map.keysSet allMod)
+                    then pure $ Map.restrictKeys allMod is
+                    else throwErrorHere (OtherError "import: cannot import undefined symbols")
                 Nothing -> pure allMod
           let qualifier = case qual of
                 Qualified q    -> ((q <> Ident "/") <>)
--- a/src/Radicle/Internal/PrimFns.hs
+++ b/src/Radicle/Internal/PrimFns.hs
@@ -716,7 +716,7 @@ purePrimFns = fromList $ allDocs $
                     else throwErrorHere
                       $ OtherError
                       $ "import: cannot import undefined symbols: "
-                        <> T.intercalate ", " (sort (fromIdent <$> Set.toList missing))
+                        <> T.intercalate ", " (fromIdent <$> Set.toAscList missing)
                 Nothing -> pure allMod
           let qualifier = case qual of
                 Qualified q    -> ((q <> Ident "/") <>)
--- a/src/Radicle/Internal/PrimFns.hs
+++ b/src/Radicle/Internal/PrimFns.hs
@@ -709,10 +709,14 @@ purePrimFns = fromList $ allDocs $
           let allMod = fromEnv e
           toImport <- case v_ of
                 Just vs -> do
-                  is <- Set.fromList <$> Protolude.toList <$> traverse isAtom vs ?? toLangError (OtherError "import: must be given a vector of symbols to import")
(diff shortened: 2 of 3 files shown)


src/Radicle/Internal/PrimFns.hs
```diff
@@ -709,8 +709,10 @@ purePrimFns = fromList $ allDocs $
           let allMod = fromEnv e
           toImport <- case v_ of
                 Just vs -> do
-                  is <- Protolude.toList <$> traverse isAtom vs ?? toLangError (OtherError "import: must be given a vector of symbols to import")
-                  pure $ Map.restrictKeys allMod (Set.fromList is)
+                  is <- Set.fromList <$> Protolude.toList <$> traverse isAtom vs ?? toLangError (OtherError "import: must be given a vector of symbols to import")
```
curiousleo: First converting to a list and then to a set is awkward, but I couldn't find any builtin to convert Seq to Set directly.

src/Radicle/Internal/PrimFns.hs
```diff
@@ -709,8 +709,10 @@ purePrimFns = fromList $ allDocs $
...
+                  if Set.null (is Set.\\ Map.keysSet allMod)
```
curiousleo: No idea if this `if ... then ... else` is idiomatic here, it's been a while since I've last written any Haskell.

src/Radicle/Internal/PrimFns.hs
```diff
@@ -709,8 +709,10 @@ purePrimFns = fromList $ allDocs $
...
+                    then pure $ Map.restrictKeys allMod is
+                    else throwErrorHere (OtherError "import: cannot import undefined symbols")
```
curiousleo: This does not tell the user which symbols were undefined - that's something that could be added.

jkarni: Maybe:

```
                  let missing = is Set.\\ Map.keysSet allMod
                  if Set.null missing
                    then pure $ Map.restrictKeys allMod is
                    else throwErrorHere
                       $ OtherError
                       $ "import: cannot import undefined symbols: "
                         <> T.intercalate ", " (fromIdent <$> Set.toList missing)
```

src/Radicle/Internal/PrimFns.hs
jkarni: I think that's fine

src/Radicle/Internal/PrimFns.hs
jkarni: Yup - looks good.

src/Radicle/Internal/PrimFns.hs
jkarni: Somehow I missed this comment:

The corresponding test (in `test/spec/Radicle/Tests.hs`, around ln 560), would be:

```
    , testCase "Importing complains if imports are undefined" $ do
        let prog = [s|(module
                        {:module 'foo :doc "" :exports '[x]}
                        (def x 1))
                      (import foo '[b a])
                      |]
        prog `failsWith` OtherError "import: cannot import undefined symbols: a, b"
```

curiousleo: Wow, thanks!

I've adjusted your suggestions slightly:
- The list of problematic imports now appears sorted in the error message to make it "more deterministic" (I think alphabetic sorts depend on environment variables related to language settings like `LC_ALL`, but I don't think that's a big issue).
- I've amended the test case to also include the case where a symbol is neither defined nor exported.

bin/rad-project
```diff
@@ -13,7 +13,7 @@
 (import prelude/cmd-parsing :unqualified)
 (import prelude/io :as 'io)
 (import prelude/io-utils
-  '[set-git-config get-git-config process-git-with-exit!]
+  '[set-git-config! get-git-config! process-git-with-exit!]
```
curiousleo: The changes in this PR surfaced these two typos. I'm a bit confused how the code in `rad-project` was able to run with the erroneous imports.

src/Radicle/Internal/PrimFns.hs
```diff
@@ -709,8 +709,14 @@ purePrimFns = fromList $ allDocs $
...
-                  is <- Protolude.toList <$> traverse isAtom vs ?? toLangError (OtherError "import: must be given a vector of symbols to import")
-                  pure $ Map.restrictKeys allMod (Set.fromList is)
+                  is <- (Set.fromList . Protolude.toList <$> traverse isAtom vs) ?? toLangError (OtherError "import: must be given a vector of symbols to import")
+                  let missing = is Set.\\ Map.keysSet allMod
+                  if Set.null missing
+                    then pure $ Map.restrictKeys allMod is
+                    else throwErrorHere
+                      $ OtherError
+                      $ "import: cannot import undefined symbols: "
+                        <> T.intercalate ", " (sort (fromIdent <$> Set.toList missing))
```
jkarni: `Set.toList` already returns the elements sorted (though you can use `Set.toAscList` to be more explicit), so no need to sort again.

bin/rad-project
jkarni: Ha. Turns out prelude already includes all of `io-utils`, which is why it was working. I can remove that import from prelude (it shouldn't really be there) in a separate PR.

src/Radicle/Internal/PrimFns.hs
curiousleo: Nice! `toAscList` it is.

bin/rad-project
curiousleo: I see!

jameshaydon:
:tada:

PR #648, opened 2019-05-29
V2 (old)

jameshaydon:
To keep track of the mergeability.


jameshaydon:
@MeBrei There was a problem with the merging of the `New Show command` PR; it brought in some random commits which is making rebasing `v2` against master quite confusing. We should fix it with a rebase.

PR #650, opened 2019-05-29
Remove reloads of error-messages module

MeBrei:
error-message.rad is already loaded by prelude. This caused its
tests to run several times.

### DIFF
--- a/rad/monadic/items.rad
+++ b/rad/monadic/items.rad
@@ -4,7 +4,6 @@
  '[summary pretty-item-view list-items pretty-state enrich-item
    verify-item-number]}
 
-(file-module! "prelude/error-messages.rad")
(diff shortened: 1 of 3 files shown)


jkarni:
While you're at it, maybe remove all of the imports in prelude related to IO? They shouldn't be there, since they're also sent to machines.

jameshaydon:
> since they're also sent to machines

No, what is sent to machines is `pure-prelude-code`, defined in `prelude/machine.rad`.
Yes, this is very confusing, and annoying to have to maintain two lists. We should come up with a solution.

jkarni:
> No, what is sent to machines is pure-prelude-code, defined in prelude/machine.rad.
Ah yeah

PR #652, opened 2019-05-31
Rad registry

MeBrei:


bin/rad-registry
```diff
@@ -0,0 +1,186 @@
...
+    (def project (items/verify-item-number n (list-projects machine) :project))
+    (put-str! (lookup :project-id project))
+    (newness/mark-read! machine n)))
+
+(def close-project
+  (fn [n machine]
+    (catch 'daemon-error
+      (do
+        (items/verify-item-number n (list-projects machine) :project)
+        (simple-edit-project! machine n {:state :closed})
```
jkarni: You seem to use this three times. Maybe make a helper function?

rad/monadic/items.rad
```diff
@@ -199,30 +248,30 @@
...
-                      "Item was not updated."
-                      (add-comment! machine item-number added-comment))))
-               _ (put-str! "No selection made."))
-        (do
-          (put-str! (pretty-table-header))
-          (map (fn [d] (put-str! (first d))) filtered-items)
-          (put-str! (summary (items-name item-flag) items)))))))
+      (match (lookup-maybe :fancy options)
+        [:just #t] (match (fzf-select-with-preview!
+                       (map first filtered-items)
```
jkarni: This feels kind of confusingly formatted. I'd say ideally in a `match` the pattern is to the lhs of the clause. Also, I think traditionally arguments are indented more than the function. 

(This will of course depend on your editor, but for me, setting the filetype to clojure makes the editor do quite sensible indentation)

rad/monadic/registry-remote.rad
```diff
@@ -0,0 +1,172 @@
...
+    (def n (project-counter :next))
+    (def i_
+      (strip-input
+       (<> i
+           {:number   n})))
+    (set-projects (@ n) i_)
+    (noncer [:use (lookup :nonce i)])
+    n))
+
+(def merge-new
```
jkarni: You can probably just use short lambdas inline, and delete this function.

rad/monadic/registry.rad
```diff
@@ -0,0 +1,125 @@
...
+    (send-code! id (find-module-file! "monadic/registry-remote.rad"))
+    (machine/send-signed-command! id radicle-registry-machine-id 'add-admin {:key owner-keys})
+    id))
+
+(def simple-project
+  "Register an open project with no labels."
+  (fn [i]
+    (def t (now!))
+    (<> {:labels      []
+         :state       :open
```
jkarni: Again, I think not having `open`/`closed` as state, but just removing from the ref, makes more sense.

rad/prelude/error-messages.rad
```diff
@@ -82,11 +82,11 @@
 
 (def uppercase-item
   (fn [item]
-    (lookup item {:issue "Issue" :patch  "Patch"})))
+    (lookup item {:issue "Issue" :patch "Patch" :project "Project"})))
 
 (def lowercase-item
   (fn [item]
-    (lookup item {:issue "issue" :patch  "patch"})))
+    (lookup item {:issue "issue" :patch "patch" :project "project"})))
```
jkarni:
```
(def lowercase-item [item]
  (drop 1 item))
```

rad/prelude/error-messages.rad
```diff
@@ -82,11 +82,11 @@
 
 (def uppercase-item
   (fn [item]
-    (lookup item {:issue "Issue" :patch  "Patch"})))
+    (lookup item {:issue "Issue" :patch "Patch" :project "Project"})))
```
jkarni:
```
(def uppercase-item [item]
  (titlecase (lowercase-item item))
```

You'll need to add some sort of pattern matching or similar functionality to implement `titlecase`

rad/prelude/io.rad
```diff
@@ -260,6 +260,16 @@ process exited normally and `[:error n]` otherwise. Example: `(process!
     (write-handle! (stderr!) prompt)
     (read-line-handle! (stdin!))))
 
+(def-rec prompt-non-empty!
+  "Same as `prompt!`, but repeats id the input empty."
```
jkarni:
```suggestion
  "Same as `prompt!`, but repeats if the input is empty."
```

rad/prelude/io.rad
```diff
@@ -260,6 +260,16 @@ process exited normally and `[:error n]` otherwise. Example: `(process!
     (write-handle! (stderr!) prompt)
     (read-line-handle! (stdin!))))
 
+(def-rec prompt-non-empty!
```
jkarni: This doesn't seem to be a very general function. Ideally the abstraction is something that takes a validator, an error message, and repeats the prompt until the validator passes.

rad/monadic/registry-remote.rad
```diff
@@ -0,0 +1,172 @@
...
+
+(def validator/project-input
+  "A project should have all the required fields, be a signed input, etc."
+  (validator/and
+   [;; NOTE that we don't check `validator/input` at this point, because some of
+    ;; the data is fixed before validation.
+    validator/time-created
+    validator/full-project]))
+
+(def edit-keys
```
jkarni: Can you explain what this does, or have a more descriptive name?

rad/monadic/items.rad
```diff
@@ -118,6 +146,24 @@
         "Comments\n"
         "--------\n\n"
         (pretty-comments cs)))
+      '{:number n
```
jkarni: It's very weird to me to have a single function that is supposed to know how to render every different item comment type. Why not different functions?

bin/rad-registry
```diff
@@ -0,0 +1,186 @@
...
+    (string-append (io/base-path!) "/registry.rad")))
+
+(def init-registry-file!
+  (fn []
+    (io/shell! (string-append "mkdir -p " (io/base-path!)) "")
+    (io/init-file-dict! (path!))))
+
+;; TODO Replace with server hosted machine
+(def default-machine-name
+  "The default registry machine."
```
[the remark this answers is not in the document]
MeBrei: It doesn't have to be a function. Changing it.

rad/monadic/items.rad
MeBrei: I agree that items.rad needs some refactoring. In this case it is convenient to have the function `pretty-item-view` handle distinguishing between the item types, otherwise there would be more matching where the function is used. 
Did you have an idea of how to split it into different functions? I can only think of an option to pass flags to the functions of items and depending on that use specific functions, but I am not sure that will improve things.

bin/rad-registry
```diff
@@ -0,0 +1,186 @@
...
+    "12D3KooWNdyJ1QpNy5RbBfjKEx1gPvRJ72tgYo1vs6jvXU8yuQN2"))
+
+(def machine-name
+  (fn []
+    (init-registry-file!)
+    (match (io/read-file-key! (path!) :registry)
+      :nothing    (default-machine-name)
+      [:just 'r]  r)))
+
+(def valid-states ["open" "closed"])
```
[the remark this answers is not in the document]
MeBrei: Will remove the `state` add commands for deleting projects.

rad/prelude/error-messages.rad
MeBrei: The wording changed. In the code `registered-machine` is used. In the UI we decided on `Radicle machine` so the flags and item names don't always match.

rad/prelude/error-messages.rad
MeBrei: (See above) The flag and item name does not match in this way anymore.

rad/monadic/items.rad
jkarni: I agree that flags are also not good (even in `pretty-table-header` it looks bad).
I don't entirely understand the "more matching" problem. There'd be a function, `pretty-issue`, that would call `pretty-issue-header`, and so on. You don't need to match on the comment (unless you want to) if you already know at call-site what shape it'll have (which you will since you'll be calling from inside a function that knows the comment shape).

MeBrei: But then you would still have to decide what function to call depending on the flag. So for example in `list-items` for each of the following function you would have to match/decide which specific one to call: `pretty-headline`, `pretty-row`, `pretty-table-header`. So would you have them refer to the specific functions depending on the flag?

jkarni: > So for example in list-items for each of the following function you would have to match/decide which specific one to call: pretty-headline, pretty-row, pretty-table-header

I don't understand this. Can you explain?

Here's the rule. If I define a new machine, with it's own set of items, and I don't have write access to this repo, can I re-use any of the functionality? Clearly, right now the answer is "no". That means the abstractions are wrong. Maybe there is no common abstraction, but probably the answer is more of a middle ground: we can't have "do everything" functions such as `pretty-item-view`, but we can have other functions that respect actual invariants. E.g.: 

```
(def prompt-comment (fn [post] ...))
```
Which takes some string `post` to put below the `---`, puts that and the line in a file, and sets you up in an editor, and gives you back the string that was inputed.

rad/prelude/io.rad
```diff
@@ -260,6 +260,18 @@ process exited normally and `[:error n]` otherwise. Example: `(process!
...
```
jkarni: The name is now outdated.

rad/prelude/io.rad
```diff
@@ -260,6 +260,18 @@ process exited normally and `[:error n]` otherwise. Example: `(process!
...
+  "Same as `prompt!`, but returns the `err-msg` and repeats the prompt if the
+  input does not pass the `validator`."
+  (fn [prompt validator err-msg]
+    (write-handle! (stderr!) prompt)
+    (def input (read-line-handle! (stdin!)))
```
jkarni: why not just use `prompt!`?

rad/monadic/items.rad
MeBrei: I agree with you. I think I am just not fully clear on which functions to abstract and which not, yet (fe `list-items`) I think the general direction is clear, but I think it exceeds this PR as it would also change issue and patch code. I can write create an issue for the refactoring.

rad/prelude/io.rad
MeBrei: Yes, sorry for that!

rad/monadic/items.rad
jkarni: ok

MeBrei: Refactoring issue: #665

rad/monadic/items.rad
MeBrei: I tried to clean it up a bit in https://github.com/radicle-dev/radicle/pull/652/commits/fe520a861557b03e595ca63cc12e4fcb5ca71af5

PR #655, opened 2019-06-04
Issues frontend

jkarni:
This gives every issues machine a frontend at the URL `http://localhost:8909/v0/machines/<machineid>/frontend/index.html`. Some caveats:

- So far only reading works (even though the plumbing is there for creating issues). This is because we need @MeBrei's changes in `v2` for writing.
- The `machineid` in the URL is that of the issues machine, so sort of annoying to figure out. I think the right approach is to have another frontend for the project, which links to the issues, but that hasn't been implemented here.

The CID is that of the build of commit 14738f8 in https://github.com/radicle-dev/rad-issues-frontend/pull/3.

### DIFF
--- a/rad/monadic/issue-remote.rad
+++ b/rad/monadic/issue-remote.rad
@@ -250,6 +250,10 @@
   "List existing issues."
   (fn [] (read-ref issues)))
 
+(def get-html
+  "The frontend."
+  (fn [] "QmQ1mvvp2kSNZwFQV2RAABdYFNEHYV3HXXEk64ZUGiKUX8"))
+
 ;; General logic
 
 (def commands
@@ -260,6 +264,7 @@
    'add-comment             add-comment
    'edit-comment            edit-comment
    'list-issues             list-issues
+   'get-html                get-html
    'add-admin               (add-admin auth)})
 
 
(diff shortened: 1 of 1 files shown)


rad/monadic/issue-remote.rad
```diff
@@ -260,6 +276,7 @@
    'add-comment             add-comment
    'edit-comment            edit-comment
    'list-issues             list-issues
+   'get-html                get-html
    'add-admin               (add-admin auth)})
```
MeBrei: Should `'set-html` also be an allowed command per default?

rad/monadic/issue-remote.rad
```diff
@@ -244,12 +244,28 @@
...
+
+(def get-html
+  "The frontend."
+  (fn [] (read-ref html)))
+
+(def set-html
+  "The frontend."
+  (fn []
+      (auth [:allowed :admin])
+      (read-ref html)))
```
MeBrei: Should this be `write-ref` or is the idea not to be able to actually change it? 
It would be helpful if `get-html` and `set-html` don't have the same doc string.

rad/monadic/issue-remote.rad
jkarni: Fixed

rad/monadic/issue-remote.rad
jkarni: Fixed.

PR #663, opened 2019-06-06
New rad repl

### LINKED ISSUE #643: rad issue list (pipe)
How come i can NOT pipe the output of `rad issue list`?

e.G.

`rad patch list | tail -n1`
or
`rad issue list 2> >(tail -n1)`

tail is ignored.

jkarni:
This makes the rad repl a properly separate executable. Advantages:

- Colors in the repl
- Output for scripts now actually goes to stdout (so fixes #643).
- Multiline input. If the parser is expecting more, the repl will continue with the same expression if you hit enter. E.g.:

```
rad> (+ 
...  3
...  2)
5
rad> ))
[repl]:1:1:
  |
1 | ))
  | ^
unexpected ')'
expecting boolean, dict, identifier, keyword, list, number, quote, short-lambda, string, or vector
```

It may also now be easier to get completion to update with new definitions, but I haven't tried to do that yet.

The PR still needs cleaning up, but since it works and we want it for the meetup, I thought I'd already open this.


jameshaydon:
@jkarni it would be nice to get this merged soon. Maybe check there weren't any additional changes to this in the demo branch.

PR #671, opened 2019-06-13
nix: remove redundant LANG attribute

asymmetric:
It's already defined in `buildStackProject`, see [0].

[0]:
https://github.com/asymmetric/nixpkgs/tree/49dc8087a20e0d742d38be5f13333a03d171006a/pkgs/development/haskell-modules/generic-stack-builder.nix#L39

### DIFF
--- a/nix/shell.nix
+++ b/nix/shell.nix
@@ -11,6 +11,5 @@ with pkgs;
 haskell.lib.buildStackProject {
   inherit ghc;
   name = "radicle";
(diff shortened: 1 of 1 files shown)


asymmetric:
Just to clarify: the link points to my fork by mistake, but the option [is set](https://github.com/NixOS/nixpkgs/blob/release-19.03/pkgs/development/haskell-modules/generic-stack-builder.nix#L39) in the `release-19.03` tag upstream too.

PR #674, opened 2019-07-11
Update rad-daemon-ipfs

juanmanual:
Installation link has moved.


jkarni:
Thanks for all these!

PR #676, opened 2019-07-11
doc: fix stack invocation in nix-shell

asymmetric:
Passing `--nix-packages zlib` errored with "You cannot have packages and
a shell-file filled at the same time in your nix-shell configuration."

The PR also documents how to install inside a nix-shell.

### DIFF
--- a/nix/README.md
+++ b/nix/README.md
@@ -26,4 +26,5 @@ And then, inside the shell:
 
 ``` sh
 stack build --system-ghc
+stack install --system-ghc :rad :radicle
 ```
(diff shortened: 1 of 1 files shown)


asymmetric:
Honestly I don't know which one is more accurate. I feel like `stack --nix build` is appropriate for when stack is available *outside* a nix-shell, as it will implicitly create one. In this case, we're already in a nix-shell, so don't know if `--nix` is the right thing to do, or `--system-ghc` (which is implied by `--nix`).

I'll add `--nix` though, as it seems to be your preference.

PR #683, opened 2019-10-09
Separate Package Boundaries

FintanH:
During my adventure into the `radicle` codebase I found there was two packages that could be extracted out into their own packages: `radicle-lang` and `radicle-repl`.

`radicle-lang` consists of only the language modules and is used by `radicle` and `radicle-repl`. The modules consist of:
```
├── Radicle
│   ├── Lang
│   │   ├── Annotation.hs
│   │   ├── Core.hs
│   │   ├── Crypto.hs
│   │   ├── Doc.hs
│   │   ├── Eval.hs
│   │   ├── Identifier.hs
│   │   ├── Internal
│   │   │   ├── Arbitrary.hs
│   │   │   ├── Orphans.hs
│   │   │   ├── Time.hs
│   │   │   └── UUID.hs
│   │   ├── Interpret.hs
│   │   ├── Json.hs
│   │   ├── Number.hs
│   │   ├── Parse.hs
│   │   ├── Pretty.hs
│   │   ├── PrimFns.hs
│   │   └── Type.hs
│   └── TH.hs
└── Radicle.hs
``` 

`radicle-repl` consists of the effects that make up the repl system. The modules that were ported over and renamed were `Radicle.Internal.Effects. and `Radicle.Internal.Effects.Capabilities`. The modules consist of:
```
└── Radicle
    ├── Repl
    │   └── Capabilities.hs
    └── Repl.hs
```

There was some minimum surgery to make things fit and to separate boundaries, but all unit tests were preserved and pass.


radicle-repl/src/Radicle/Repl/Capabilities.hs
```diff
@@ -2,14 +2,14 @@
...
-module Radicle.Internal.Effects.Capabilities where
+module Radicle.Repl.Capabilities where
 
 import           Protolude
 
 import qualified Data.ByteString as BS
 import           Data.Text.Prettyprint.Doc (PageWidth)
 import           Data.Time
-import           Paths_radicle
+import           Paths_radicle_repl
```
FintanH: Not so sure about this change. I think this was for a call to `getDataDir` which I believe is actually the directory above. Need to look into that

radicle-lang/package.yaml
```diff
@@ -0,0 +1,94 @@
+author:              "Author name here"
```
kim: :eyes:

radicle-repl/src/Radicle/Repl/Capabilities.hs
kim: True, `findModule` will try to resolve to the top-level `rad` dir. Perhaps this could be moved to the `Radicle.Internal.CLI` module, which isn't so much about CLI but about paths (so could be renamed to ... `Paths`)?

kim: Oh wait, does that create circular dependencies?

PR #684, opened 2019-10-14
Separate Package Boundaries (Take 2 Electric Bugaloo)

FintanH:
A non-forked version of #683 due to CI reasons.

During my adventure into the `radicle` codebase I found there was two packages that could be extracted out into their own packages: `radicle-lang` and `radicle-repl`.

`radicle-lang` consists of only the language modules and is used by `radicle` and `radicle-repl`. The modules consist of:
```
├── Radicle
│   ├── Lang
│   │   ├── Annotation.hs
│   │   ├── Core.hs
│   │   ├── Crypto.hs
│   │   ├── Doc.hs
│   │   ├── Eval.hs
│   │   ├── Identifier.hs
│   │   ├── Internal
│   │   │   ├── Arbitrary.hs
│   │   │   ├── Orphans.hs
│   │   │   ├── Time.hs
│   │   │   └── UUID.hs
│   │   ├── Interpret.hs
│   │   ├── Json.hs
│   │   ├── Number.hs
│   │   ├── Parse.hs
│   │   ├── Pretty.hs
│   │   ├── PrimFns.hs
│   │   └── Type.hs
│   └── TH.hs
└── Radicle.hs
``` 

`radicle-repl` consists of the effects that make up the repl system. The modules that were ported over and renamed were `Radicle.Internal.Effects. and `Radicle.Internal.Effects.Capabilities`. The modules consist of:
```
└── Radicle
    ├── Repl
    │   └── Capabilities.hs
    └── Repl.hs
```

There was some minimum surgery to make things fit and to separate boundaries, but all unit tests were preserved and pass.


FintanH:
> Just a nit: I'm not sure it's a good idea to check the *.cabal files in when using hpack. Is the output deterministic?

I agree actually. I should have asked earlier, but I did it this way because I assumed `radicle.cabal` was checked-in in the same fashion. How was that updated? I'm also unaware of how `hpack` packages get hosted on Hackage because afaik they need the cabal file in the project 🤔 Guess I should do some looking into that for my own sake

kim:
> `radicle.cabal` was checked-in in the same fashion.

It's actually `.gitignore`'d at the top-level.

> how `hpack` packages get hosted on Hackage

`stack` honours `package.yaml` files, and invokes `hpack` automatically. So, when you do `stack sdist`, it'll package the `*.cabal` file (**not** the `package.yaml`).

FintanH:
Thanks for the explanation! I removed them and updated the ignore to ignore _all_ cabal files :)

PR #688, opened 2019-10-28
CI: Set up Buildkite

kim:
E2E tests and packaging remain in cloudbuild for now, due to IPFS and docker compose shenanigans. As we're moving away from IPFS, we'll sunset cloudbuild wholesale.

Fixes radicle-dev/cocopoco#7


asymmetric:
> we're moving away from IPFS

@kim is there more information about this?

